PR #2, opened 2017-05-30
fix(deps): Add missing express and open dependencies for the cli

thewtex:


jourdain:
oups... Thanks for fixing it.

jourdain:
We may want to use shelljs to do the copy instead of my inline command as it may not work on windows.

thewtex:
> We may want to use shelljs to do the copy instead of my inline command as it may not work on windows.

Done :-): https://github.com/Kitware/itk-vtk-image-viewer/pull/3

PR #17, opened 2017-10-10
Transfer function widget

thewtex:
@jourdain please take a look

itk-vtk-image-viewer-transfer-function-snapshot

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -66,8 +66,11 @@ const VIEWER_MAPPING = {
     mapper.setSliceAtFocalPoint(true);
 
     const actor = vtkImageSlice.newInstance();
-    actor.getProperty().setColorWindow(255);
-    actor.getProperty().setColorLevel(127);
+    const dataArray = data.image.getPointData().getScalars();
+    const dataRange = dataArray.getRange();
+    const window = dataRange[1] - dataRange[0];
+    actor.getProperty().setColorWindow(window);
+    actor.getProperty().setColorLevel(dataRange[0] + (window / 2.0));
     actor.setMapper(mapper);
     const iStyle = vtkInteractorStyleImage.newInstance();
     iStyle.setInteractionMode('IMAGE_SLICING');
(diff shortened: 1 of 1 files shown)


src/viewer.js
```diff
@@ -121,12 +110,69 @@ function createViewer(container, data) {
...
+    handleColor: 'rgb(50, 150, 50)',
+    buttonDisableFillColor: 'rgba(255, 255, 255, 0.5)',
+    buttonDisableStrokeColor: 'rgba(0, 0, 0, 0.5)',
+    buttonStrokeColor: 'rgba(0, 0, 0, 1)',
+    buttonFillColor: 'rgba(255, 255, 255, 1)',
+    strokeWidth: 2,
+    activeStrokeWidth: 3,
+    buttonStrokeWidth: 1.5,
+    handleWidth: 3,
+    iconSize: 20, // Can be 0 if you want to remove buttons (dblClick for (+) / rightClick for (-))
```
jourdain: The buttons are kind of optional if you know (+) double click (-) right click.

src/viewer.js
```diff
@@ -121,12 +110,69 @@ function createViewer(container, data) {
...
   const renderWindow = fullScreenRenderer.getRenderWindow();
 
+  const dataArray = data.image.getPointData().getScalars();
+  if (!dataArray) {
+    window.alert('No data array available in dataset');
+  }
+  const dataRange = dataArray.getRange();
+
+  const lookupTable = vtkColorTransferFunction.newInstance();
+  // TODO: Make colormap selectable
```
jourdain: Do you want me to try to improve the UI a little bit with draggable window and preset selection?

src/viewer.js
thewtex: I think it is nice to have them explicitly there -- hard to guess that behavior otherwise.

jourdain: ok no worries.. Without them you can have slicker UI.

src/viewer.js
thewtex: Yes, please!

Also, I was trying to figure out a way to limit the renderWindow size in pixels, for our container purposes, so it does not hammer the system and become very slow on high DPI screens. Any ideas?

jourdain: I'll give it a try tomorrow.

Are you saying the slowness depend on the resolution of the 3D window? or the piecewise widget?

thewtex: Thanks!

The 3D window -- fullscreen on a 4k display requests many pixels.

jourdain: I'll see if I can add some options to limit the resolution like we do in ParaViewWeb.

PR #27, opened 2018-01-10
chore(Webpack): Webpack externals config shim is no longer required w…

thewtex:
…ith itk.js:7.0.0

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -31,7 +31,4 @@ module.exports = {
       sourcePath,
     ],
   },
-  externals: {
-    config: '{}',
-  },
 };


jourdain:
Are you using an object named `config` that you expect to be available on the global namespace?

If so, you need to keep the file the way it was.

thewtex:
The use of [`config`](https://www.npmjs.com/package/config) was removed with itk-js 7.0.0. In practice, in Node, the same relative paths are used in the package. For the browser, webpack provides equivalent functionality. This removal make it easier to use when using webpack.

PR #38, opened 2018-01-29
Proxy view

thewtex:
Refactor to use the 3D viewer to use the ProxyView.

Next steps are to create the 2D view states.

### DIFF
--- a/src/createViewer.js
+++ b/src/createViewer.js
@@ -8,7 +8,7 @@ const STYLE_CONTAINER = {
   width: '100%',
   height: '100%',
   minHeight: '200px',
-  minWidth: '200px',
+  minWidth: '450px',
   margin: '0',
   padding: '0',
   top: '0',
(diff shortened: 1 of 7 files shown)


src/createViewer.js
```diff
@@ -0,0 +1,80 @@
...
+  view.setContainer(container);
+  view.resize();
+
+  userInterface.createToggleUI(rootContainer);
+
+  let imageSource = null;
+  let lookupTable = null;
+  let piecewiseFunction = null;
+  if (image) {
+    imageSource = proxyManager.createProxy('Sources', 'TrivialProducer');
```
jourdain: Other way to write it to skip line 56.

`imageSource = proxyManager.createProxy('Sources', 'TrivialProducer', { name: 'Image' });`

src/createViewer.js
```diff
@@ -0,0 +1,80 @@
...
+    imageSource.setInputData(image);
+    imageSource.setName('Image');
+
+    proxyManager.createRepresentationInAllViews(imageSource);
+    const representation = proxyManager.getRepresentation(imageSource, view);
+    lookupTable = representation.getLookupTableProxy();
```
jourdain: You can get the LookupTable/PiecewiseFunction proxy directly on the ProxyManager with the Array name.

```
const dataArray = image.getPointData().getScalars();
lookupTable = proxyManager.getLookupTable(dataArray.getName());
lookupTable = proxyManager.getPiecewiseFunction(dataArray.getName());
```

see https://github.com/Kitware/vtk-js/blob/master/Sources/Proxy/Core/ProxyManager/properties.js#L82-L104

thewtex:
@jourdain Thanks for the tips and guidance! Added.

PR #47, opened 2018-02-08
Ijk

thewtex:


thewtex:
After chatting with @agirault, it sounds like better support for orientation of the slices is coming into vtk.js, so I will wait until that is in, then revisit this and #48

thewtex:
Superceded by #56 -- we now do slicing in world coordinates, like a boss. :construction_worker_man:

PR #49, opened 2018-02-09
World position picking

thewtex:
Thanks for the great ideas, Alexis!

@jourdain @agirault the data probe sphere position only gets updated when I click the mouse instead of just moving the mouse -- any ideas why?

@agirault The MouseWheel event callback was disabled because `callData.position` was `undefined` -- a bug?


src/ItkVtkViewProxy.js
```diff
@@ -144,9 +140,9 @@ function ItkVtkViewProxy(publicAPI, model) {
   model.interactor.onMouseMove((event) => {
     updateAnnotations(event);
   });
-  model.interactor.onMouseWheel((event) => {
-    updateAnnotations(event);
-  });
+  // model.interactor.onMouseWheel((event) => {
+  //  updateAnnotations(event);
+  // });
```
agirault: the mouse wheel event that is forwarded does not have information regarding position. If you want to reuse a position information from a previous event (mousemove) you'll need to explicitly store it.

thewtex: Thanks for the note. Removed.

thewtex:
New version uses: https://github.com/Kitware/vtk-js/pull/575 h/t @agirault

PR #54, opened 2018-02-14
Jupyter

thewtex:


agirault:
Any way that the `css-element-queries` dependency for faster resizing detection could be part of vtk-js to always have views resize more responsively? cc: @jourdain

jourdain:
Nope, that is not the responsibility of vtk.js. Just the application using it and managing its layout.

PR #56, opened 2018-02-23
VTK performance updates and non-axis-aligned slicing in physical coordinates

thewtex:


thewtex:
@jourdain @agirault it works! Thank you!! :pray: :1st_place_medal:

PR #59, opened 2018-02-23
feat(ImageUI): Limit number of slider value characters to 6

thewtex:


src/userInterface/createImageUI.js
```diff
@@ -271,8 +272,8 @@ function createPlaneIndexSliders(
   function updateZSlice() {
     const value = Number(zSliceElement.value);
     volumeRepresentation.setZSlice(value);
-    const valueString = String(zSliceElement.value);
-    const padLength = valueString.length < 4 ? 4 - valueString.length : 0;
+    const valueString = String(zSliceElement.value).substring(0, numberOfValueChars);
+    const padLength = valueString.length < numberOfValueChars ? numberOfValueChars - valueString.length : 0;
```
agirault: just fiy @thewtex : https://www.w3schools.com/jsref/jsref_tofixed.asp

thewtex: Thanks for the tip.

Here we are looking to use a fixed number of significant digits / characters as opposed to fixed decimal points, but there may definitely be a better way to do it.

agirault: gotcha

PR #62, opened 2018-03-02
WIP: feat(ImageUI): Update window / level with mouse drag

thewtex:
CC: @girault @jourdain


agirault:
I have it in WIP also here: https://github.com/Kitware/itk-vtk-image-viewer/tree/improve-interactions

agirault:
Could you show me where you link the transfer function gaussian update to the window level update? And I can update my PR based on what I have and yours.

thewtex:
@agirault that is fantastic! Closing this one, then.

After digging into the PiecewiseGaussianWidget a bit more, it seems like adjusting the `selectedGaussian` is not a good idea because it is only selected when the mouse is inside the widget.

Instead, this PR: adds `.getGaussians()`:

  https://github.com/Kitware/vtk-js/pull/622

We could adjust the Gaussian `width`, `position`, and `height` at index 0, i.e. `gaussians[0]` with different mouse drags to adjust the window, level, and opacity, respectively. Then call `.setGaussians(gaussians)`.

PR #64, opened 2018-03-07
Improve interactions

thewtex:


src/proxyManagerConfiguration.js
```diff
@@ -1,13 +1,31 @@
...
 
+const commonInteractor = [
+  { type: 'pan', options: { button: 3 } }, // Pan on Right button drag
+  { type: 'pan', options: { button: 1, shift: true } }, // Pan on Shift + Left button drag
+  { type: 'zoom', options: { button: 1, control: true } }, // Zoom on Ctrl + Left button drag
+  { type: 'zoom', options: { dragEnabled: false, scrollEnabled: true } }, // Zoom on scroll
+];
+
+const interactorStyle3D = [
+  Array.from(commonInteractor).concat([
```
jourdain: Concat create a new array, so you don't need `Array.from`.

src/proxyManagerConfiguration.js
```diff
@@ -51,14 +65,17 @@ const proxyManagerConfiguration = {
           viewUp: [0, 0, 1], // Z+ (S)
           useParallelRendering: false,
         },
+        props: {
+          presetToInteractor3D: interactorStyle3D,
+          presetToInteractor2D: interactorStyle2D,
+        },
```
jourdain: Just to be sure. Do you have a `setPresetToInteractor2/3D(...)` method? If so, then you are good.

src/proxyManagerConfiguration.js
thewtex: :+1: I'll remove `Array.from`.

src/proxyManagerConfiguration.js
thewtex: Yes, @agirault set these, and I found that it works in practice :-). They are defined in vtk.js here: 

  https://github.com/Kitware/vtk-js/blob/41b0da325781d0a3bbe082a22927ec9fbea53bef/Sources/Proxy/Core/ViewProxy/index.js#L77-L118

In general, it would be nice to have documentation for the proxy manager configuration.

thewtex:
Thanks for the the review, @jourdain !

PR #81, opened 2018-07-06
Vtk js bump

thewtex:


src/createViewer.js
```diff
@@ -125,6 +125,8 @@ const createViewer = (
   });
   proxyManager.renderAllViews();
 
+  setTimeout(view.resetCamera, 1);
```
thewtex: @jourdain is this the correct solution?

PR #85, opened 2018-07-24
Transfer function widgets viewer pipeline

thewtex:


karma.conf.js
```diff
@@ -81,7 +81,7 @@ module.exports = function init(config) {
     customLaunchers: {
       ChromeHeadlessNoSandbox: {
         base: 'ChromeHeadless',
-        flags: ['--no-sandbox'],
+        flags: ['--no-sandbox', '--ignore-gpu-blacklist'],
```
thewtex: @jourdain this seems to do the trick for WebGL CI tests :fireworks:

jourdain: This is great... I was waiting for chrome to fully work headless with webgl.

PR #124, opened 2018-12-07
chore(Dependencies): Bump vtk.js to 8.3.0

thewtex:
@jourdain @floryst setting hyperdrive to warp speed to bump vtk.js from 7.8.2 to 8.3.0. :man_astronaut: I seem to be almost there, but the default volume orientation seems to have changed:

selection_054

Any ideas why the change and how the old orientation can be restored?


jourdain:
Looking at the change log here is the list of possible commit that may have changed that behavior:
- https://github.com/kitware/vtk-js/commit/b26f7ef
- https://github.com/kitware/vtk-js/commit/d8b1537
- https://github.com/kitware/vtk-js/commit/ba24cf8

But I don't know, it is just a guess based on the commit messages...

thewtex:
Thanks for the pointers @jourdain.

I looked around, but did not find anything. But it is fine -- I added orientation in the view initialization.

PR #126, opened 2018-12-12
ci(semantic-release): Specify repositoryUrl

thewtex:
For publishing. The GIT_PUBLISH_URL is setup in .travis.yml. Use for
repositoryUrl if set.

This helps semantic-release pick up the correct publish URL.

As documented:
https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/configuration.md#repositoryurl

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "ITK/VTK Image Viewer",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


thewtex:
@jourdain thanks for the review and tip! :taco: 

Pushing suggested change...

PR #137, opened 2019-05-03
Fix tooltips blocking interactions

Fumon:
Fixes a specific, annoying issue where the "Volume Spacing Distance" and "Gradient Opacity" tooltips pop up and block interactions with the windowing canvas. This fix makes the tooltips only popup when hovering over the relevant icons.
Tested on Firefox 66.0.3 x64 on Windows 10


thewtex:
Fantastic. Thank you @Fumon !

PR #167, opened 2019-08-31
Color map range zoom

thewtex:


thewtex:
:ok_hand: @jourdain 


localhost_8080 (1)

PR #196, opened 2019-12-03
Initial multi-channel volume

thewtex:
@martinken @jourdain adding in multi-component transfer function widget and multi-component volume rendering support :+1: 

InitialMultiComponent

More todo...


jourdain:
Looks cool

PR #249, opened 2020-04-02
WIP: Replace lut pwf select with tabs

scottwittenburg:
This is coming along.  Still needed:

- [x] Better styling of tabs
- [x] text of labels better contrast
- [x] Improve look/feel of visibility checkboxes
- [x] Hook up component visibility to volume rendering
- [ ] with 2 components, second one doesn't seem to be registered until you click the tab
- [x] if only one component, don't show the tabs (and don't allow turning that comp off)

### DIFF
--- a/src/UserInterface/Image/createComponentSelector.js
+++ b/src/UserInterface/Image/createComponentSelector.js
@@ -1,16 +1,21 @@
 import { reaction, action } from 'mobx'
 
+import updateSliceProperties from '../../Rendering/updateSliceProperties'
+import updateVolumeProperties from '../../Rendering/updateVolumeProperties'
 import style from '../ItkVtkViewer.module.css'
 
 function createComponentSelector(store, imageUIGroup) {
   const viewerDOMId = store.id
 
-  const componentSelector = document.createElement('select')
+  const componentSelector = document.createElement('div')
   componentSelector.setAttribute('class', style.selector)
   componentSelector.id = `${viewerDOMId}-componentSelector`
 
   const componentRow = document.createElement('div')
   componentRow.setAttribute('class', style.uiRow)
+  // This row needs custom bottom padding, to aid in the illusion
+  // that it's the tabbed portion of a tabbed pane
+  componentRow.setAttribute('style', 'padding-bottom: 0px;')
   componentRow.className += ` ${viewerDOMId}-volumeComponents ${viewerDOMId}-toggle`
 
   function updateAvailableComponents() {
@@ -25,8 +30,20 @@ function createComponentSelector(store, imageUIGroup) {
       .fill(undefined)
       .map((_, ii) => ii)
       .map(
-        component =>
-          `<option value="${component}" >Component ${component}</option>`
+        (idx, component) =>
+          `<input name="tabs" type="radio" id="tab-${component}" ${
+            idx === 0 ? 'checked="checked"' : ''
+          } class="${
+            style.componentTab
+          }" data-component-index="${component}"/><label for="tab-${component}" class="${
+            style.compTabLabel
+          }">${component}<input type="checkbox" ${
+            store.imageUI.componentVisibilities[idx] > 0.0
+              ? 'checked="checked"'
+              : ''
+          } class="${
+            style.componentVisibility
+          }" data-component-index="${component}"\></label>`
       )
       .join('')
     componentSelector.value = 0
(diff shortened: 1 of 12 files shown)


scottwittenburg:
@thewtex More work to do here, but the faked tabs don't look too bad given that they're actually radio buttons ;-), and the checkboxes actually work to turn on/off the contributions of the various components, but only for the slices.

src/UserInterface/Image/createComponentSelector.js
```diff
@@ -45,9 +59,21 @@ function createComponentSelector(store, imageUIGroup) {
   componentSelector.addEventListener(
     'change',
     action(event => {
+      console.log('changed component')
       event.preventDefault()
       event.stopPropagation()
-      store.imageUI.selectedComponentIndex = Number(event.target.value)
+      const selIdx = Number(event.target.getAttribute('data-componentIndex'))
```
thewtex, line 65: Could be cleaner here by using `dataset` exposed via [HTML5 Data Attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes).

scottwittenburg: Thanks for pointing that out, it would be cleaner.  Will do.

scottwittenburg: Thanks for pointing that out, I wasn't aware.  Will do.

scottwittenburg:
@thewtex This could go in now if you like, and I can create a new issue for the remaining unchecked item in the description.  I'm struggling to figure out why we see the behavior, and it could make sense as a separate bug anyway.

src/UserInterface/index.js
```diff
@@ -8,6 +8,7 @@ import createPointSetsUI from './createPointSetsUI'
 import emptyContainer from './emptyContainer'
 import getRootContainer from './getRootContainer'
 import checkForWebGL from './checkForWebGL'
+import { updateSliceActors, updateVolumeProps } from './updateActors'
```
thewtex, line 11: Let's start organizing the rendering-related logic in `src/Rendering/`, so these functions can be exposed in `src/Rendering/index.js` instead of `src/UserInterface`.

src/UserInterface/updateActors.js
```diff
@@ -0,0 +1,46 @@
+
+function updateSliceActors(store) {
```
thewtex, line 2: `updateSliceActors` -> `updateSliceProperties` for consistency with the other function and since it mostly updates the properties.

src/UserInterface/updateActors.js
```diff
@@ -0,0 +1,46 @@
...
+        const visibility = store.imageUI.componentVisibilities[component];
+        actorProp.setRGBTransferFunction(component, lutProxy.getLookupTable());
+        actorProp.setPiecewiseFunction(component, pwfProxy.getPiecewiseFunction());
+        actorProp.setComponentWeight(component, visibility);
+      }
+    })
+  }
+}
+
+function updateVolumeProps(store) {
```
thewtex, line 21: `updateVolumeProps` -> `updateVolumeProperties` -> more specific, consistent, and not confusable with props, as in items in the scene.

src/ViewerStore.js
```diff
@@ -55,6 +55,7 @@ class ImageUIStore {
 
   lookupTableProxies = []
   piecewiseFunctionProxies = []
+  componentVisibilities = []
```
thewtex, line 58: We want to make these `@observable`, then we can react to their changes.

src/UserInterface/Image/createComponentSelector.js
```diff
@@ -47,19 +63,22 @@ function createComponentSelector(store, imageUIGroup) {
...
       event.stopPropagation()
-      store.imageUI.selectedComponentIndex = Number(event.target.value)
+      const selIdx = Number(event.target.dataset.componentIndex)
+      if (event.target.type === 'radio') {
+        store.imageUI.selectedComponentIndex = selIdx
+      } else if (event.target.type === 'checkbox') {
+        const visibility = event.target.checked ? 1.0 : 0.0
+        store.imageUI.componentVisibilities[selIdx] = visibility
+        updateSliceActors(store)
+        updateVolumeProps(store)
```
thewtex, line 73: We want these to calls to go into a reaction for `store.imageUI.componentVisibilities.slice()`. This way, they will always get called when the component visibilities are changes, which could happen in the user interface or from other sources, i.e. an API call that originates from Python activities.

src/UserInterface/updateActors.js
```diff
@@ -0,0 +1,46 @@
...
+        volumeProperty.setRGBTransferFunction(component, lut)
+        volumeProperty.setScalarOpacity(component, piecewiseFunction)
+        volumeProperty.setComponentWeight(component, visibility)
+      }
+    })
+  }
+}
+
+export {
+  updateSliceActors,
```
thewtex, line 44: Exporting one function per module makes them easy to find.

src/ViewerStore.js
thewtex: This makes it easier to have multiple reactions for a given state change.

scottwittenburg:
@thewtex I fixed the color/contrast on the selected tab as we discussed earlier today, and I think I addressed all the issues you raised in review here.  Thanks for providing feedback!

PR #251, opened 2020-04-03
build(pre-commit-hook): Run prettier as a pre-commit Git hook

thewtex:
Option 1 from: https://prettier.io/docs/en/precommit.html


thewtex:
Should make formatting-to-consistent-style a non-issue :-)

Issue #257, opened 2020-04-08
Wrong color map for common RGB images


oeway:
It seems the latest viewer cannot show regular JPEG/png images properly anymore, the default color map looks wired.

scottwittenburg:
Hmmm, we're always setting `independentComponents = true` on the vtkImageMapper with the recent changes.  I wonder if showing JPEG/png images requires it to be `false`, thus invoking the previous shaders.

thewtex:
Yes, if the data is 2D / Uint8Array / 3 or 4 component, we should assume `independentComponents = false`

oeway:
It seems this issue comes back after recent PRs, now if we load a jpeg image for example, the color map is not correct by default. @thewtex

thewtex:
@oeway is there a sample image file I should try? The PNG's and JPEG's I tried look good to me.

oeway:
Hey, i tried a few on my mobile and some of them works indeed, but photos took from a camera doesn't work. Will try to provide some example image later.

I also noticed that in ImageJ.JS, https://ij.imjoy.io if I open a sample image via File-> samples -> flybrain, then  open it in Itk-vtk-viewer (in the plugin menu), the 3D image I got is reddish, while the previous version shows the correct RGB color in 3D.

thewtex:
> Hey, i tried a few on my mobile and some of them works indeed, but photos took from a camera doesn't work. Will try to provide some example image later.

Great, thanks.

> I also noticed that in ImageJ.JS, https://ij.imjoy.io if I open a sample image via File-> samples -> flybrain, then open it in Itk-vtk-viewer (in the plugin menu), the 3D image I got is reddish, while the previous version shows the correct RGB color in 3D.

This sounds familiar during recent refactoring, but it may have been resolved -- is it still the case with current `master`?

oeway:
@thewtex Here is one example with the RGB color image: https://images.proteinatlas.org/61448/1319_C10_2_blue_red_green.jpg  If I load it into the viewer, it [becomes reddish](https://kitware.github.io/itk-vtk-viewer/app/?fileToLoad=https://images.proteinatlas.org/61448/1319_C10_2_blue_red_green.jpg).

For 3D images, I confirm that the color issue has been fixed in ImageJ.JS. However, there seems some additional noise on the rendered images compared to the previous version: 
[...]

PR #258, opened 2020-04-10
WIP: fix(Viewer): Make sure 2D images are rendered as before

scottwittenburg:
Sets `independentComponents = false` under conditions described in #257.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10190,8 +10190,7 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "aproba": {
               "version": "1.2.0",
@@ -10212,14 +10211,12 @@
             "balanced-match": {
               "version": "1.0.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "brace-expansion": {
               "version": "1.1.11",
               "bundled": true,
               "dev": true,
-              "optional": true,
               "requires": {
                 "balanced-match": "^1.0.0",
                 "concat-map": "0.0.1"
@@ -10234,20 +10231,17 @@
             "code-point-at": {
               "version": "1.1.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "concat-map": {
               "version": "0.0.1",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "console-control-strings": {
               "version": "1.1.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "core-util-is": {
               "version": "1.0.2",
@@ -10364,8 +10358,7 @@
             "inherits": {
               "version": "2.0.4",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "ini": {
               "version": "1.3.5",
@@ -10377,7 +10370,6 @@
               "version": "1.0.0",
               "bundled": true,
               "dev": true,
-              "optional": true,
               "requires": {
                 "number-is-nan": "^1.0.0"
               }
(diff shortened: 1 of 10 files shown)


scottwittenburg:
@thewtex When showing a 2D RGB image, should we still show the component selector?

oeway:
Would it make sense to still use independent components, but set default color maps for displaying RGB and alpha? 

Also, I think it would be nice if we can choose different modes programmatically with publicAPI.

thewtex:
> (or more specifically, when we set independentComponents = false), should we still show the component selector?

The component selector should be hidden in this case.

thewtex:
> Would it make sense to still use independent components, but set default color maps for displaying RGB and alpha?
> 
> Also, I think it would be nice if we can choose different modes programmatically with publicAPI.

Selection of independent may be possible -- we would have to constrain it to the cases when it is possible to use both independent or dependent components.  Dependent components is likely a better default for the 2D RGB case since that is standard way those images are visualized and the extra complexity may be distracting.

scottwittenburg:
> > (or more specifically, when we set independentComponents = false), should we still show the component selector?
> 
> The component selector should be hidden in this case.

But we should still show the grayscale colormap editor and that histogram panel?

thewtex:
> But we should still show the grayscale colormap editor and that histogram panel?

Previously, the default piecewise function in this case was set to be effectively all 1's. It would ideally be removed, but the window-level slider bar and histogram are still needed.

scottwittenburg:
> Would it make sense to still use independent components, but set default color maps for displaying RGB and alpha?

@oeway Do you have an image you could share that would let us explore whether using independent components with RGB images might make sense or not?

> Also, I think it would be nice if we can choose different modes programmatically with publicAPI.

I'm not sure I understand your intention here.  Which modes you want to choose programmatically?  What API should expose those options?

scottwittenburg:
> > But we should still show the grayscale colormap editor and that histogram panel?
> 
> Previously, the default piecewise function in this case was set to be effectively all 1's. It would ideally be removed, but the window-level slider bar and histogram are still needed.

Just checking what you mean here.  I understand you want to keep the histogram in 2D mode, that's clear.

Re the default (all 1's) piecewise function:  It seems to be used in rendering the image/volume, so we should keep it, but do you want to remove the ability to edit it, delete it, add more pwfs, etc?

Re the window-level slider bar: I don't see this.  Am I just looking at a dataset that doesn't result in it getting added to the view?

scottwittenburg:
> > Would it make sense to still use independent components, but set default color maps for displaying RGB and alpha?
> 
> @oeway Do you have an image you could share that would let us explore whether using independent components with RGB images might make sense or not?
> 
> > Also, I think it would be nice if we can choose different modes programmatically with publicAPI.
> 
> I'm not sure I understand your intention here. Which modes you want to choose programmatically? What API should expose those options?

So I've just noticed the publicAPI methods in the bottom of `createViewer.js`.  Are you embedding the viewer in some other application of your own and using those methods to control it?  In that light, maybe I do understand your request.  You're probably asking to expose the `independentComponents` on the volume/image property objects?

oeway:
Hi, thanks for working on this. [Here](https://images.proteinatlas.org/41082/721_D4_2_blue_red_green.jpg) is one image I would like to display. On the one hand, I would like to see the image appears in RGB mode and be able to adjust brightness for all components jointly, on the other hand, I would like to adjust each component independently to balance the brightness for each channel.

For the publicAPI, yes, I am using it in other applications and would like to able to access the independentComponents programmatically.

thewtex:
> Re the default (all 1's) piecewise function: It seems to be used in rendering the image/volume, so we should keep it, but do you want to remove the ability to edit it, delete it, add more pwfs, etc?

Since we have the ability to define a piecewise transfer function in the multi-component case :sun_with_face: , we could make it editable only in the case of multiple-components or 3D -- in the case of 2D scalar data it does have an effect. If we can fix it, great, but it is not a big priority.

> Re the window-level slider bar: I don't see this.

image

It shows up as blue when mousing over the transfer function editor.

> You're probably asking to expose the independentComponents on the volume/image property objects?

:+1:

scottwittenburg:
@thewtex I think this has everything we talked about for a short-term set of fixes, please take a look when you get a chance.

PR #261, opened 2020-04-15
fix(Volume): Fix volume initialization issue

scottwittenburg:


scottwittenburg:
@thewtex I tried to create another `store.imageUI.image` reaction that would run after all the others (there are 3 or 4 of them throughout the codebase), and while I was able to achieve that, it didn't fix the problem.  Instead I just had to move this block of code to the earliest possible spot in the code that fixed the problem.  But I guess there's still something I don't understand about the way the reactions are registered and run.

We can iterate on this if you want, I just wanted to stick in a placeholder.

scottwittenburg:
> Thanks, @scottwittenburg !
> 
> Can this go either:
> 
> * At the end of transfer function widget initialization in _createTransferFunctionWidget.js_, or

I have tried putting it there, but it didn't solve the problem.

> * After line 228 in _createViewer.js_ (after we have set up the image UI

I tried there also, no luck.

Perhaps there are other reactions that happen as a result of things changing in the 3-4 `store.imageUI.image` reactions, and it's only after all those cascading reactions occur that iterating through the selected components does what we expect.  That would explain why adding this code into a final `store.imageUI.image` reaction doesn't help, but adding it directly after the line that triggers the `store.imageUI.image` reactions does the trick.

> ?
> 
> With the goal to make it local to where it is needed (state initialization related to the transfer function widget), and to cover other ways of initializing the image, e.g. from the multiscaleManager or `publicAPI.setImage`.

Yes, I definitely agree with that goal.  I wasn't specifically aware of those other places where the image was initialized (so thanks for pointing them out), but I figured it was likely possible.

If I can find all the reactions that need to happen before this code executes (in order for the code to solve the problem), that will help me figure out how to place this code into the `createTransferFunctionWidget.js` module.

thewtex:
@scottwittenburg while digging in, I may have found it in #262 -- please take a look.

Issue #266, opened 2020-04-23
Embedded viewer no more loading


kerautret:
Congratulation for your excellent viewer that is very nice to display online results and demonstration. Thanks a lot for your valuable work!

I detected something wrong by using the viewer like described here: https://kitware.github.io/itk-vtk-viewer/docs/embeddedViewer.html
It was perfectly working using old version less or equal to 9.18.0 (not tested all versions) like here:
http://ipol-geometry.loria.fr/SourceCodeDemosWorkshop/dist/testEmbeded9.18.0.html
But on using the current version (or same with 9.19.0) the viewer is not loading my test image (.nii)
http://ipol-geometry.loria.fr/SourceCodeDemosWorkshop/dist/testEmbededCurrent.html
http://ipol-geometry.loria.fr/SourceCodeDemosWorkshop/dist/testEmbeded9.19.0.html
I always get the same view : 
<img width="1119" alt="Capture d’écran 2020-04-23 à 14 48 08" src="https://user-images.githubusercontent.com/772865/80101267-780ae500-8571-11ea-8f33-cf68321e2617.png">

It looks like that the viewer is trying to read the first letter of the field:   data-url
<img width="1231" alt="Capture d’écran 2020-04-23 à 14 45 58" src="https://user-images.githubusercontent.com/772865/80101060-27938780-8571-11ea-9d5a-5e24a933a780.png">

By the way did there exist a way do display two image volume in one viewer ? (or mixing with a .obj file) ?

scottwittenburg:
@kerautret Thanks for reporting this.  We recently added support for passing both an intensity image as well as a labelmap via the `fileToLoad` url parameter, and I suspect this may have broken the file loading for the embedded use case.  I'll look into it, and hopefully have a fix out quickly.

scottwittenburg:
@kerautret I don't have a quick/easy way to test embedding the viewer, but I think the fix I just submitted should take of the problem.

kerautret:
@scottwittenburg many thanks for your reply. I test it ;)

kerautret:
Excellent the fix is perfectly working nice 🎉

kerautret:
by the way did there exists a way to load several files ? or an mesh file?

kerautret:
@scottwittenburg and by the way the load of several objects is also working, I see that différents objects can be added with , .

PR #267, opened 2020-04-23
fix(Viewer): Fix broken url handling in embedded case

scottwittenburg:
Adding support for passing two comma-separated urls in the `fileToLoad` url parameter broke the url handling when embedding the viewer in another application and using the `data-url` attribute to specify the location of the data.  This fixes that, extending the support for providing comma-separated urls in the embedded use case.


scottwittenburg:
Hopefully fixes #266

kerautret:
Thanks a lot this work: 
http://ipol-geometry.loria.fr/SourceCodeDemosWorkshop/itk-vtk-viewer/dist/testEmbeded.html

PR #270, opened 2020-04-25
fix(createGeometryColorChooser): Default geometry color is red

thewtex:
Has a good contrast with the viridis colormap used for images.

### DIFF
--- a/src/UserInterface/Geometries/createGeometryColorChooser.js
+++ b/src/UserInterface/Geometries/createGeometryColorChooser.js
@@ -8,7 +8,7 @@ function createGeometryColorChooser(store, geometryColorRow) {
   geometryColorInput.setAttribute('type', 'color')
   geometryColorInput.id = `${store.id}-geometryColorInput`
 
-  const defaultGeometryColor = '#ffffff'
+  const defaultGeometryColor = '#ff0000'
 
   reaction(
     () => {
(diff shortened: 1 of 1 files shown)


kerautret:
Thanks a lot that works perfectly ;) 👍🏻

Capture d’écran 2020-04-26 à 10 51 15

kerautret:
Perhaps the same edit can be done for the alpha selection ? For general use the default value can be let to 1.0 but it could a way to change it in a custom version.
https://github.com/Kitware/itk-vtk-viewer/blob/master/src/UserInterface/Geometries/createGeometryOpacitySlider.js

thewtex:
> Perhaps the same edit can be done for the alpha selection ? For general use the default value can be let to 1.0 but it could a way to change it in a custom version.
https://github.com/Kitware/itk-vtk-viewer/blob/master/src/UserInterface/Geometries/createGeometryOpacitySlider.js

Yes, could you please create a PR?

kerautret:
> > Perhaps the same edit can be done for the alpha selection ? For general use the default value can be let to 1.0 but it could a way to change it in a custom version.
> > https://github.com/Kitware/itk-vtk-viewer/blob/master/src/UserInterface/Geometries/createGeometryOpacitySlider.js
> 
> Yes, could you please create a PR?

Yes I do it :) Thanks for this new version

PR #275, opened 2020-05-01
Multiscale labelmap

thewtex:


src/createViewer.js
```diff
@@ -46,9 +46,6 @@ const createViewer = (
   UserInterface.applyContainerStyle(rootContainer, store, viewerStyle)
 
   let updatingImage = false
-  if (!!labelMap) {
-    store.imageUI.labelMap = labelMap
```
scottwittenburg, line 50: @thewtex Moving this initialization is what broke the behavior of the interpolation button, as it checked that property on the store to know whether interpolation should be on/off and the button enabled/disabled.  

That is done as a part of `createMainUI`, and I would have thought that other things within the creation of the main ui might need to know whether or not we have a `labelMap`.  But I could be wrong.  Any thoughts as to whether I could move this initialization back before the `createMainUI`?

PR #276, opened 2020-05-04
Fix issue affecting labemaps with many labels

scottwittenburg:
This PR fixes the regression with the interpolation button being enabled, even when we we have a labelmap, as well as some issues we saw last week with lots of labels seemingly getting assigned the same color, even though mouse picking could distinguish the labeled regions.

fix-issue-with-many-labels-opt

### DIFF
--- a/src/Rendering/createLabelMapRendering.js
+++ b/src/Rendering/createLabelMapRendering.js
@@ -2,6 +2,15 @@ import vtkLookupTableProxy from 'vtk.js/Sources/Proxy/Core/LookupTableProxy'
 import vtkPiecewiseFunction from 'vtk.js/Sources/Common/DataModel/PiecewiseFunction'
 import applyCategoricalColorToLookupTableProxy from './../UserInterface/applyCategoricalColorToLookupTableProxy'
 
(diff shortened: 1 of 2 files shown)


scottwittenburg:
@thewtex It seems to me like this has greatly improved the volume rendering of labelmaps with many labels, so we may no longer need to get into the opacity mode `fractional` vs `proportional` stuff we discussed on Friday.  But maybe you want to try out the branch and let me know what you think.

thewtex:
> @thewtex It seems to me like this has greatly improved the volume rendering of labelmaps with many labels,

@scottwittenburg woah, yes, much better! :clap: :1st_place_medal: We are seeing all the colors now, no artifacts :100: 

> we may no longer need to get into the opacity mode fractional vs proportional stuff we discussed on Friday. But maybe you want to try out the branch and let me know what you think.

This should still be useful because, as seen in this example where we have a lot of colors, but there is just a lot of clutter and no control over the overall opacity; this will provide a way to a) emphasize the content of interest, b) define the content of interest, c) define a way to control the overall opacity (we could do this in other ways, but a) and b) are even more useful).

PR #278, opened 2020-05-05
fix(LabelMaps): Support loading and displaying annotations

scottwittenburg:
Supports loading annotations via url parameter `annotations`.  Example annotations file can be found [here](https://thewtex.github.io/allen-ccf-itk-vtk-zarr/allen_ontology_50.json).

### DIFF
--- a/src/ItkVtkViewProxy.js
+++ b/src/ItkVtkViewProxy.js
@@ -10,7 +10,7 @@ import vtkWidgetManager from 'vtk.js/Sources/Widgets/Core/WidgetManager'
 import * as vtkMath from 'vtk.js/Sources/Common/Core/Math'
 
 const CursorCornerAnnotation =
-  '<table class="corner-annotation" style="margin-left: 0;"><tr><td style="margin-left: auto; margin-right: 0;">Index:</td><td>${iIndex},</td><td>${jIndex},</td><td>${kIndex}</td></tr><tr><td style="margin-left: auto; margin-right: 0;">Position:</td><td>${xPosition},</td><td>${yPosition},</td><td>${zPosition}</td></tr><tr><td style="margin-left: auto; margin-right: 0;"">Value:</td><td>${value}</td></tr></table>'
+  '<table class="corner-annotation" style="margin-left: 0;"><tr><td style="margin-left: auto; margin-right: 0;">Index:</td><td>${iIndex},</td><td>${jIndex},</td><td>${kIndex}</td></tr><tr><td style="margin-left: auto; margin-right: 0;">Position:</td><td>${xPosition},</td><td>${yPosition},</td><td>${zPosition}</td></tr><tr><td style="margin-left: auto; margin-right: 0;"">Value:</td><td>${value}</td></tr><tr ${annotationStyle}><td style="margin-left: auto; margin-right: 0;">Annotation:</td><td colspan="3">${annotation}</td></tr></table>'
 
 const { vtkErrorMacro } = macro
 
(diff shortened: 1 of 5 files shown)


scottwittenburg:
@thewtex, while this is basically working end-to-end, I think we'll need to iterate on it a little.  For example:

1. I can't remember if the label values correspond to the index in, e.g. `allen_ontology_50.json`, or if it's the `id` values.
2. I show the 'acronym` field, rather than the `name` field, because of how long the `name` seems to be.
3. This does not yet support drag-and-drop annotations files.

src/ItkVtkViewProxy.js
```diff
@@ -93,6 +93,14 @@ function ItkVtkViewProxy(publicAPI, model) {
     }
   }
 
+  function getAnnotationText(value) {
+    const labelValue = value[model.labelIndex]
```
scottwittenburg, line 97: Should this function live here and the `annotationMap` and `labelIndex` provided to this thing's `model`?  Or should we just provide this proxy class with a method that can be called to retrieve the annotation text?

thewtex:
Looking good, thanks @scottwittenburg! I will take a closer look later on today.

src/ItkVtkViewProxy.js
```diff
@@ -10,7 +10,7 @@ import vtkWidgetManager from 'vtk.js/Sources/Widgets/Core/WidgetManager'
 import * as vtkMath from 'vtk.js/Sources/Common/Core/Math'
 
 const CursorCornerAnnotation =
-  '<table class="corner-annotation" style="margin-left: 0;"><tr><td style="margin-left: auto; margin-right: 0;">Index:</td><td>${iIndex},</td><td>${jIndex},</td><td>${kIndex}</td></tr><tr><td style="margin-left: auto; margin-right: 0;">Position:</td><td>${xPosition},</td><td>${yPosition},</td><td>${zPosition}</td></tr><tr><td style="margin-left: auto; margin-right: 0;"">Value:</td><td>${value}</td></tr></table>'
+  '<table class="corner-annotation" style="margin-left: 0;"><tr><td style="margin-left: auto; margin-right: 0;">Index:</td><td>${iIndex},</td><td>${jIndex},</td><td>${kIndex}</td></tr><tr><td style="margin-left: auto; margin-right: 0;">Position:</td><td>${xPosition},</td><td>${yPosition},</td><td>${zPosition}</td></tr><tr><td style="margin-left: auto; margin-right: 0;"">Value:</td><td>${value}</td></tr><tr><td style="margin-left: auto; margin-right: 0;">Annotation:</td><td>${annotation}</td><td></td><td></td></tr></table>'
```
thewtex, line 13: Can the *Annotation* row's presence be conditional on the presence of a label map?

src/ItkVtkViewProxy.js
thewtex: This looks good to me :+1:

src/ItkVtkViewProxy.js
```diff
@@ -93,6 +93,14 @@ function ItkVtkViewProxy(publicAPI, model) {
     }
   }
 
+  function getAnnotationText(value) {
+    const labelValue = value[model.labelIndex]
+    if (model.annotationMap != null && model.annotationMap.has(labelValue)) {
+      return model.annotationMap.get(labelValue)
+    }
+    return 'Unknown'
```
thewtex, line 101: `labelValue` string is preferred.

src/createViewer.js
```diff
@@ -23,6 +23,7 @@ const createViewer = (
     multiscaleImage,
     labelMap,
     multiscaleLabelMap,
+    annotations,
```
thewtex, line 26: -> `labelMapAnnotations` since geometries could have lookup tables.

src/processFiles.js
```diff
@@ -176,6 +187,14 @@ export const readFiles = async ({
       webWorker.terminate()
       labelMapData = vtkITKHelper.convertItkToVtkImage(itkImage)
     }
+    let annotationData = null
+    if (!!annotations) {
+      annotationData = new Map(
+        annotations
+          .filter(item => !!item)
+          .map(({ acronym }, idx) => [idx, acronym])
```
thewtex, line 195: -> factor out for generality.

scottwittenburg:
@thewtex Mind taking another look when you get a chance?  I think I addressed all the concerns you raised, and if we can get it merged, I would like to make use of some of this work in #282.

Thanks!

PR #282, opened 2020-05-06
fix(Viewer): Make labelmap component proportional in volume rendering

scottwittenburg:
Needs this [change](https://github.com/Kitware/vtk-js/pull/1434) from `vtk.js`.

proportional-component-optimize

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,12 +55,12 @@
       }
     },
     "@babel/generator": {
-      "version": "7.9.5",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.9.5.tgz",
-      "integrity": "sha512-GbNIxVB3ZJe3tLeDm1HSn2AhuD/mVcyLDpgtLXa5tplmWrJdF/elxB56XNqCuD6szyNkDi6wuoKXln3QeBmCHQ==",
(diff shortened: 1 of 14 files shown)


scottwittenburg:
@thewtex This is still missing that extra multiplier you described, and isn't tested other than with the zarr-ized ccf.  So let's iterate on it a bit before we decide?

scottwittenburg:
@thewtex I couldn't get everything done tonight, but here's a video showing the latest:

per-label-weighting

thewtex:
@scottwittenburg I am going to start integrating at least the click selection functionality into Jupyter -- would you please factor out the *clickCallback* improvements here into a separate commit that can be merged while working out the opacity details? In the process it would be best to have a *subscribe* interface so we can have two callbacks registered -- updating the selected label map and sending the last clicked data down to Python.

scottwittenburg:
@thewtex I think this is ready for review.  As it turned out, enhancing the label `select` wasn't easy and straightforward, so I wanted to defer that until we come up with a better UI for managing those per-label weights.  I think I addressed the other items we discussed friday.

Next PR will update the event subscription infrastructure and expose the API we'll need to embed the viewer as a value widget in a Jupyter notebook.

Issue #260, opened 2020-04-13
Re-enable CI browser rendering tests


thewtex:
The Docker browser rendering tests were disabled because they are failing on CI systems in #259 even though they pass locally.

PR #283, opened 2020-05-11
ci(GraphicalTests): Re-enable

thewtex:
Tests were failing with false positives due to Firefox 75. Firefox 76 is
now out.


thewtex:
Closes #260

thewtex:
@floryst running into

```
> karma start ./karma.conf.js --browsers Chrome,Firefox "--browsers" "Firefox"

sh: 1: karma: not found
```

but just for Node 12. Does this ring a bell? :bell: :sound:

floryst:
Yeah, that's due to [needle having issues right now with last node 12 version](https://github.com/tomas/needle/issues/312). You can install `request` as a dev dependency as a workaround.

thewtex:
@floryst :1st_place_medal: thanks!

PR #284, opened 2020-05-14
fix(Viewer): Expose picked values via click subscription

scottwittenburg:


scottwittenburg:
@thewtex There's no other subscription management functionality in the store, but I put this one there so we would have it available to any of our components, as well as making it available to embedding applications.  Let me know if a cleaner approach comes to mind.

oeway:
Nice! I am thinking that would a even more generic event bus make sense? The viewer can expose a standard [EventEmitter](https://github.com/EventEmitter2/EventEmitter2/blob/master/README.md) interface(with functions including on, once, off, emit etc.), one can subscribe to many different events. Not only pick value, but also view change, camera position change etc.

thewtex:
@scottwittenburg nice!

@oeway thanks for the suggestion! 

@scottwittenburg could you please look into providing subscriptions and exposing the `EventEmitter` interface via the `eventemitter2` package?

scottwittenburg:
> @scottwittenburg could you please look into providing subscriptions and exposing the `EventEmitter` interface via the `eventemitter2` package?

Yes, will do. 👍

thewtex:
I noticed `eventemitter3` is available -- I am guessing it is better than `eventemitter2` ;-).

https://www.npmjs.com/package/eventemitter3

thewtex:
Required API's for Juptyer

- [ ] `publicAPI.setLabelMap`

thewtex:
Superceded by #289

PR #286, opened 2020-05-21
feat(GradientOpacitySlider): Adjust on a log scale

thewtex:
Gradient magnitudes for natural images are likely to have a power-law
distribution. To provide linear scale to the user adjusting gradient
opacity with the slider, make gradient opacity maximum and minimum steps
on a log scale.


thewtex:
@martinken @jourdain would you like this pushed upstream into vtk.js?

jourdain:
@martinken ?

martinken:
Log works for me and makes sense. Didn't check the code too carefully but if it feels good on the GUI tests with a few datatsets then +1

PR #290, opened 2020-05-28
Label map updates

thewtex:


thewtex:
@scottwittenburg some tweaks while working on the Jupyter widget.

src/ItkVtkViewProxy.js
```diff
@@ -133,7 +133,12 @@ function ItkVtkViewProxy(publicAPI, model) {
           xPosition: String(worldPosition[0]).substring(0, 4),
           yPosition: String(worldPosition[1]).substring(0, 4),
           zPosition: String(worldPosition[2]).substring(0, 4),
-          value,
+          value:
+            model.labelIndex === null
+              ? value
```
scottwittenburg, line 138: After this change, `value` is no longer defined, causing a console error when there is no label map.

src/createViewer.js
```diff
@@ -116,27 +114,17 @@ const createViewer = (
...
-                  store.imageUI.labelMapWeights[store.imageUI.selectedLabel]
-                if (currentWeight > 0.5) {
-                  store.imageUI.labelMapWeights[
-                    store.imageUI.selectedLabel
-                  ] = 0.0
-                } else {
-                  store.imageUI.labelMapWeights[
-                    store.imageUI.selectedLabel
-                  ] = 1.0
-                }
```
scottwittenburg, line 136: I preferred to see the weight of each label when clicking it, instead of immediately changing it to 1.  But this is behavior, together with the meaning of the `All` selection, is really up to what you prefer.

src/ItkVtkViewProxy.js
thewtex: Great catch  -- this should be `fusedValue`

src/createViewer.js
thewtex: Makes sense. This is a refinement to move towards a preference of efficient interaction, i.e. just one click.

PR #293, opened 2020-06-02
fix(viewer): Various rendering/ui improvements

scottwittenburg:
Detect when we have only a labelmap and make adjustments to how user-
specified weights (from the labelmap piecewise function slider) are
mapped to the actual pwf.  Also, adjust weight->pwf mapping when we
are showing slice compared to volume, as we always want some context
in slice mode (so we can see segments to adjust pwf for).

Decouple the component selection checkboxes from the labelmap weight
slider, as those do not interact well together.  After this change,
the component visibility state consists of a boolen indicating whether
the component is shown or hidden, as well as float specifying the
weight.

Depends on this [change](https://github.com/Kitware/vtk-js/pull/1473) in `vtk.js`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,21 +5,21 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.8.3",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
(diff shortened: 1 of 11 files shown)


scottwittenburg:
Improvement in rendering when we have only labelmap:

just_labelmap-optimize

scottwittenburg:
Improvement when we have at least one intensity and a labelmap:

intensity_plus_labelmap-optimize

scottwittenburg:
@thewtex Let me know what you think.  Instead of just multiplying the fragment alpha by the proportional component pwf value, I used the proportional component weight to adjust that scaling factor.  I thought this made sense because when you have set the labelmap component weight really small, I feel that piecewise function should have a smaller effect on the fragment opacity.

PR #294, opened 2020-06-03
WIP: Fixes for many component images

scottwittenburg:
Add support for when images have more components that we can visualize at one time.

This is getting close, some remaining items:

- [ ] need to recompute the max gradient opacity value each time we re-fuse image
- [ ] When we hit a checkbox and we have to re-fuse the images, the lighting is turned back on but the checkbox isn't kept in sync
- [ ] Show component names if available (add a store datastructure, a way to populate it, and a way to load data for it into application)
- [ ] Every time the `fusedImage` changes, we go back through `createImageRendering` again, which trashes and recreates the lookup table and piecewise function proxies, and picks new default presets.  This causes us to lose any changes to colormaps we have made.  The gaussians are also changed from what we may have set, but they're not back to the initial defaults.  If I'm looking at a transfer function on a component where I picked a new colormap, then I also see a mismatch between what's drawn on top of the histogram and what's in the color selector (color selector has the replaced color, the original default, while on top of the histogram I see the new colormap I had picked but which got replaced).  I just need to look at a different one, and then come back to see the state sync'd up again.
- [ ] (Needs validation) When we have a component hidden, but it's not the most recently touched component (call this guy "unlucky"), when we show another component that requires re-fusing, then suddenly "unlucky" will be shown instead of hidden, even though I didn't click it's checkbox.
- [ ] (Needs validation) I sometimes see that old bug (after replacing a component) where the render view changes just from clicking on a different component tab
- [ ] Generally needs to be exercised with a lot of different datasets to try and find any other issues.

### DIFF
--- a/src/Rendering/createImageRendering.js
+++ b/src/Rendering/createImageRendering.js
@@ -7,24 +7,32 @@ import updateSliceProperties from './updateSliceProperties'
 
 function createImageRendering(store, use2D) {
   const numberOfComponents = store.imageUI.numberOfComponents
+  const totalComponents = store.imageUI.totalIntensityComponents
 
-  store.imageUI.lookupTableProxies = new Array(numberOfComponents)
-  store.imageUI.piecewiseFunctionProxies = new Array(numberOfComponents)
+  store.imageUI.lookupTableProxies = new Array(totalComponents)
+  store.imageUI.piecewiseFunctionProxies = new Array(totalComponents)
   const initialComponentVisibilities = []
-  for (let i = 0; i < numberOfComponents; i++) {
+  for (let i = 0; i < totalComponents; i++) {
+    const visible =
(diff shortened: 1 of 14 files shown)


scottwittenburg:
@thewtex This still has a lot of issues, but I wanted to get the ball rolling.  I'll start trying to get the issues characterized in a checklist in the PR description tomorrow.

oeway:
Thanks for working on this!

thewtex:
@scottwittenburg are updates planned on this or should we merge after a rebase? I am looking to perform refactoring to support @oeway 's requested layer support, etc :-)

scottwittenburg:
@thewtex I've been planning to work on updates to this for the last week and a half 😉 but just haven't been able to squeeze it in.  It's pretty close, but there are a few issues left to tackle, which I've added as todo items in the description.

Issue #299, opened 2020-06-09
PolyData scalar colors incorrectly shown


rubenverhack:
Thank you for the development of this great tool!

I'm new to VTK, but I was implementing a Python script that reads two VTk files, combines them using an AppendFilter and writes them to a file. 

I have noticed that if I render the output file using the following script. The colors are rendered correctly. However, if I open the file using itk-vtk-viewer, then the whole object is a single color.

There might be an issue with how the file was created, so I added the script for combining the two vtk objects as well.
[object-.zip](https://github.com/Kitware/itk-vtk-viewer/files/4752924/object-.zip)

~~~
import vtk

def read_vtk(path):
    reader = vtk.vtkPolyDataReader()
    reader.SetFileName(path)
    reader.ReadAllVectorsOn()
    reader.ReadAllScalarsOn()
    reader.Update()

    return reader
output = read_vtk('output.vtk')

# Create a mapper and actor
mapper = vtk.vtkPolyDataMapper()
# mapper.ScalarVisibilityOn()
# mapper.SetInputConnection(cleanFilter.GetOutputPort())
mapper.SetInputConnection(output.GetOutputPort())
mapper.SetColorModeToDirectScalars()

actor = vtk.vtkActor()
actor.SetMapper(mapper)

# Create a renderer, render window, and interactor
renderer = vtk.vtkRenderer()
renderWindow = vtk.vtkRenderWindow()
renderWindow.AddRenderer(renderer)
renderWindowInteractor = vtk.vtkRenderWindowInteractor()
renderWindowInteractor.SetRenderWindow(renderWindow)

# Add the actors to the scene
renderer.AddActor(actor)
renderer.SetBackground(.3, .2, .1) #  Background color dark red

# Render and interact
renderWindow.Render()
renderWindowInteractor.Start()
~~~

Output in itk-vtk-viewer:
![image](https://user-images.githubusercontent.com/28840027/84164243-95890500-aa72-11ea-8564-67e87a500acb.png)

Output in renderer:
![image](https://user-images.githubusercontent.com/28840027/84164342-b2253d00-aa72-11ea-808e-7577417c86b0.png)

[output.zip](https://github.com/Kitware/itk-vtk-viewer/files/4752882/output.zip)

Script to combine the two VTK files:
~~~
import vtk

def read_vtk(path):
    reader = vtk.vtkPolyDataReader()
    reader.SetFileName(path)
    reader.ReadAllVectorsOn()
[...]

PR #300, opened 2020-06-10
fix(Geometries): Support direct unsigned char scalar colors

thewtex:
The vtk.js AbstractRepresentationProxy sets the Mapper ColorMode to
MAP_SCALARS if scalars and a lookup table proxy are present, but we always
have a lookup table proxy. Set ColorModeToDefault on the Mapper and high
lookup table GUI elements with 3-component unsigned char scalars.


thewtex:
Closes #299

PR #301, opened 2020-06-10
Expose more viewer features via api

scottwittenburg:
Exposes `opacityGaussiansChanged` and `componentVisibilitiesChanged` events, as well as api to set/get those properties.

### DIFF
--- a/src/Rendering/updateComponentPiecewiseFunction.js
+++ b/src/Rendering/updateComponentPiecewiseFunction.js
@@ -1,0 +1,19 @@
+function updateComponentPiecewiseFunction(store, component) {
+  const lookupTableProxy = store.imageUI.lookupTableProxies[component]
+  const lookupTable = lookupTableProxy.getLookupTable()
+  const transferFunctionWidget = store.imageUI.transferFunctionWidget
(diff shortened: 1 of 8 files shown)


scottwittenburg:
@thewtex I thought I'd get this up so you can take a look while I'm exposing a few more of the things we talked about.  It required more re-working to expose the gaussians than I expected, mostly the `set` functionality, but it seems like it's working properly now, and there's a WIP commit demonstrating it.

thewtex:
@scottwittenburg nice, thanks! The piecewise Gaussians will be awesome to have.

Are you working on `labelMapBlend` or `backgroundColor`? If not, I'll pick them up.

scottwittenburg:
At the moment I'm exposing api for component visibilities.  That data structure contains the non-labelmap weights as well, which change with the labelmap blend, but I thought that should be separate api (or really just inferred from the labelmap blend).

But let me know if you agree.

thewtex:
> At the moment I'm exposing api for component visibilities. That data structure contains the non-labelmap weights as well, which change with the labelmap blend, but I thought that should be separate api (or really just inferred from the labelmap blend).

> But let me know if you agree.

Yes, :+1:

PR #304, opened 2020-06-11
Expose background color and labelmap blend

scottwittenburg:
Adds a button to the top row which allows toggling between a set of pre-defined background colors.  The `publicAPI` already exposes a way to set the background color to an arbitrary value, but this PR makes sure that gets applied to the scene.  

This PR also exposes ability to set/get the `labelMapBlend`, a scalar value indicating how much we see the labelmap color map vs the other intensity components.  Higher values of `labelMapBlend` mean we see more of the labelmap colormap, while lower values mean we see more of the other intensity components.

This PR also emits new events when either of the above properties is changed.


src/UserInterface/Main/createBackgroundColorButton.js
```diff
@@ -0,0 +1,28 @@
...
+function createBackgroundColorButton(store, mainUIRow) {
+  const viewerDOMId = store.id
+  const bgColorButton = document.createElement('div')
+  bgColorButton.innerHTML = `<input id="${viewerDOMId}-bgColorButton" type="checkbox" class="${style.toggleInput}" checked><label itk-vtk-tooltip itk-vtk-tooltip-top itk-vtk-tooltip-content="Toggle Background Color" class="${style.bgColorButton} ${style.toggleButton}" for="${viewerDOMId}-bgColorButton">${selectColorIcon}</label>`
+  const bgColorButtonLabel = bgColorButton.children[1]
+  applyContrastSensitiveStyle(store, 'invertibleButton', bgColorButtonLabel)
+  bgColorButton.addEventListener('change', event => {
+    event.preventDefault()
+    event.stopPropagation()
+    console.log('changed')
```
thewtex, line 16: rm

src/UserInterface/Main/createBackgroundColorButton.js
```diff
@@ -0,0 +1,28 @@
...
+    toggleBackgroundColor(store)
+  })
+  bgColorButton.addEventListener('click', event => {
+    event.preventDefault()
+    event.stopPropagation()
+    console.log('clicked')
```
thewtex, line 22: rm

src/UserInterface/toggleBackgroundColor.js
```diff
@@ -0,0 +1,11 @@
+function toggleBackgroundColor(store) {
+  console.log('dang')
```
thewtex, line 2: rm

scottwittenburg:
> @scottwittenburg nice! A few debug statements noted inline.

Yes, I was just testing you.  You passed the test.

thewtex:
Dang, good test :-D

PR #319, opened 2020-06-24
fix(SliceRendering): Fix color out of range issues for 2+ components

scottwittenburg:
Should fix the color out-of-range issues we have seen with 2 or more components.

### DIFF
--- a/src/Rendering/applyPiecewiseFunctionOpacities.js
+++ b/src/Rendering/applyPiecewiseFunctionOpacities.js
@@ -1,0 +1,17 @@
+import filterAndProcessOpacityNodes from './filterAndProcessOpacityNodes'
+
+function applyPiecewiseFunctionOpacities(store, component) {
+  const transferFunctionWidget = store.imageUI.transferFunctionWidget
+  const opacityNodes = transferFunctionWidget.getOpacityNodes()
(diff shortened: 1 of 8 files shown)


src/Rendering/applyPiecewiseFunctionOpacities.js
```diff
@@ -0,0 +1,43 @@
...
+    let sum = 0.0
+    let count = 0
+    for (let j = i - halfWindow; j <= i + halfWindow; j++) {
+      if (j > 0 && j < filteredNodes.length) {
+        sum += filteredNodes[j].y
+        count += 1
+      }
+    }
+    filteredNodes[i].y = sum / count
+  }
```
thewtex, line 24: For efficiency, can we do this loop for the starting and ending nodes (i.e. up to `halfWindow`, then something like:

```
const running = Array(kernelSize)
const sum = 0.0
for (let i = 0; i < kernelSize; i++) {
  running[i] = filteredNodes[i].y / kernelSize
  sum += running[i]
}

for (let i = halfWindowSize; i < filteredNodes.length -halfWindowSize; i++) {
  filteredNodes[i].y = sum
  sum -= running.shift()
  const next = filteredNodes[+1i].y / kernelSize
  running.push(next)
  sum += next
}
```

?

PR #329, opened 2020-07-27
Scale axes

thewtex:


thewtex:
@floryst this is a bit of hack of the widget framework, but loving the SVG overlay :heart_eyes: .

PR #340, opened 2020-10-02
feat(ImJoy): Add plugin interface

thewtex:


dist/index.html
```diff
@@ -42,11 +45,62 @@
...
-      var viewer = itkVtkViewer.processURLParameters(container, { fullscreen: true });
+     async function setupViewerForImJoy(container) {
+      const imjoyRPC = await imjoyLoader.loadImJoyRPC({
+        api_version: '0.2.3'
+      })
+      const api = await imjoyRPC.setupRPC({
+        name: 'itk-vtk-viewer',
+        version: '9.23.x',
+        description: '2D / 3D web image, mesh, and point set viewer using itk.js and vtk.js ',
+        type: 'rpc-window',
```
oeway, line 56: Could you please also add a `docs` key that point to a url (e.g. a web page or a markdown file) with the documentation for the ImJoy api?

dist/index.html
```diff
@@ -42,11 +45,62 @@
...
+      api.registerCodec({
+        name: 'ndarray',
+        decoder: itkVtkViewer.utils.ndarrayToItkImage
+      })
+      api.export({
+        setup() {
+          api.log("itk-vtk-viewer loaded successfully.")
+        },
+        async run(ctx) {
+          if (ctx.data && ctx.data.image_array) {
```
oeway, line 71: I used `image_array` as the key for the image input here, this might not be the best choice, maybe think a bit more to see if we can come up with another key?  Maybe simply `image`? Also think about other keys for `pointSets` and `geometries`?

dist/index.html
```diff
@@ -42,11 +45,62 @@
...
+            itkVtkViewer.createViewerFromLocalFiles(container)
+          }
+        },
+        async imshow(image_array) {
+          const vtkImage = itkVtkViewer.utils.vtkITKHelper.convertItkToVtkImage(image_array)
+          const dims = vtkImage.getDimensions()
+          const is2D = dims.length === 2 || (dims.length === 3 && dims[2] === 1)
+          itkVtkViewer.createViewer(container, {
+            image: vtkImage,
+            pointSets: null,
```
oeway, line 83: Similar to `imshow`, would be nice to add api for `pointSets` and `geometries`, what do you think?

dist/index.html
thewtex: Sounds good -- I'll add a `docs` page at https://kitware.github.io/itk-vtk-viewer/docs/imjoy.html.

dist/index.html
```diff
@@ -42,11 +45,62 @@
...
     <script>
-      var container = document.querySelector('.content');
-      var viewer = itkVtkViewer.processURLParameters(container, { fullscreen: true });
+     async function setupViewerForImJoy(container) {
+      const imjoyRPC = await imjoyLoader.loadImJoyRPC({
+        api_version: '0.2.3'
+      })
+      const api = await imjoyRPC.setupRPC({
+        name: 'itk-vtk-viewer',
+        version: '9.23.x',
```
thewtex, line 54: @oeway we will get `version` from *package.json*.

dist/index.html
thewtex: Yes, `image` is better -- I'll change it to that. :+1:  for `pointSets` and `geometries` in the future.

dist/index.html
thewtex: Yes, I'll add the `pointSets` and `geometries` API functions after we have the layers functionality implemented :-).

dist/index.html
oeway: Yes, that will be great!

dist/index.html
oeway: Great! You mean you will eventually support layers from the viewer?
Do you want to also change `imshow` to something else then? to make it easier to be consistent with `showPointSets` later.

Should we decide 2D or 3D for the user? `dims.length === 2 || (dims.length === 3 && dims[2] === 1)` might not be the best choice.

oeway:
BTW, we can also use the ImJoy badges somewhere, see here: https://imjoy.io/docs/#/development?id=add-an-imjoy-badge

dist/index.html
thewtex: > Do you want to also change imshow to something else then? to make it easier to be consistent with showPointSets later.

Yes, we can use `addImage` to go with `addGeometry`, `addPointSet`.

> Should we decide 2D or 3D for the user? dims.length === 2 || (dims.length === 3 && dims[2] === 1) might not be the best choice.

Yes, there is ambiguity with `ndarray` -- `itk/Image` provides an opportunity to explicitly define the dimension, so we will support that, too.

thewtex:
> BTW, we can also use the ImJoy badges somewhere, see here: https://imjoy.io/docs/#/development?id=add-an-imjoy-badge

:star2: :rocket: added to the doc page, doc index page, and README.

dist/index.html
oeway: Sounds good!

So I guess the idea is if we call `addImage` twice, will we get two layers, correct?

For now, would be nice to have a way to update the image without recreate the entire viewer (e.g. `updateImage` or `setImage`), to support applications such as a live view of the microscope: [Binder](https://mybinder.org/v2/gh/imjoy-team/micro-manager-imjoy/master?filepath=Micro-Manager-ImJoy-Tutorial.ipynb)

thewtex: > So I guess the idea is if we call addImage twice, will we get two layers, correct?

Yes. At least in the near future, we can only support a single image, so other layers will be made invisible. We can support multiple geometries and point sets.

> For now, would be nice to have a way to update the image without recreate the entire viewer (e.g. updateImage or setImage), to support applications such as a live view of the microscope: ￼

:heart: !

Cool, we will add a method like that, too :-)

test/imjoyTest.js
```diff
@@ -0,0 +1,125 @@
...
+  })
+  await imjoy.start({ workspace: 'default' })
+  console.log('ImJoy started')
+
+  const plugin = await imjoy.pm.reloadPluginRecursively({
+    uri: 'http://localhost:9876/base/dist/index.html',
+  })
+  console.log(plugin)
+  await plugin.api.run({ config: {}, data: {} })
+  console.log(plugin.api.addImage)
```
thewtex, line 85: @oeway I am trying to set up a testing harness here.

`plugin.api.addImage` is currently `undefined` -- what am I doing wrong?

To reproduce:

```
# Checkout the branch
npm ci
npm run build
npm run test:debug
# Click the "Debug" button in the browser window
# Open the console
# Refresh the page
```

oeway: I am on it.

oeway: Ok, I think the issue is the plugin object you get is not the actual window instance, but a proxy plugin to the window plugin. To connect to the actual window, we need to use `createWindow`. This behavior might need to be improved, it will be more intuitive to just return the window plugin.
```js
  const proxyPlugin = await imjoy.pm.reloadPluginRecursively({
    uri: 'http://localhost:9876/base/dist/index.html',
  })
  console.log(proxyPlugin)
  // instantiate a viewer, the returned object is the api directly
  const viewer = await imjoy.pm.createWindow(null, {type: "itk-vtk-viewer"})
  console.log(viewer.addImage)
```

You can also do the following (recommended), which will give you directly the viewer object.
```js
  // const proxyPlugin = await imjoy.pm.reloadPluginRecursively({
  //   uri: 'http://localhost:9876/base/dist/index.html',
  // })
  const viewer = await imjoy.pm.createWindow(null, {src: 'http://localhost:9876/base/dist/index.html'})
  console.log(viewer.addImage)

```

thewtex: @oeway verified! Thank you so much! :1st_place_medal: :star2:

package.json
```diff
@@ -36,6 +36,7 @@
     "babel-preset-mobx": "^2.0.0",
     "copy-webpack-plugin": "^4.5.1",
     "husky": "^4.2.3",
+    "imjoy-core": "^0.13.36",
```
oeway, line 39: FYI: the latest version is now `0.13.37`

thewtex:
@oeway your review would be appreciated!

Starting with `setImage` for the API, since it is implemented, serves current needs, and it should not change.

dist/index.html
```diff
@@ -35,18 +36,77 @@
...
+          if (this.viewer === null) {
+            this.viewer = await itkVtkViewer.createViewer(container, {
+              image: vtkImage,
+              pointSets: null,
+              geometries: null,
+              use2D: is2D,
+              rotate: false
+            })
+          } else {
+            await this.viewer.setImage(vtkImage)
```
oeway, line 92: Will this be an issue if we pass another image with different size, dimension (2D vs 3D) to the viewer? If not we might need to check if the new image has the same dimension as the last one.

oeway (approved):
Looks great!

I added some suggestion for the docs to clarify some additional options and add basic examples. It would be also great if we can add for executable examples, e.g. binder, colab notebooks, but that can be addressed in a following PR. Furthermore, we can think about how to best support other data types (xarray, ome-zarr etc) with optional compression.

dist/index.html
```diff
@@ -35,18 +36,77 @@
...
+      const imjoyRPC = await imjoyLoader.loadImJoyRPC({
+        api_version: '0.2.3'
+      })
+      const version = itkVtkViewer.version
+      const api = await imjoyRPC.setupRPC({
+        name: 'itk-vtk-viewer',
+        version,
+        description: '2D / 3D web image, mesh, and point set viewer using itk.js and vtk.js ',
+        type: 'rpc-window',
+        docs: 'https://kitware.github.io/itk-vtk-viewer/docs/imjoy.html',
```
oeway, line 58: Please also consider adding the following keys:
 * `authors`: a list of author names
 * `license`: name of the license
 * `labels`: e.g. ["visualization"]
 * `icon`: url to an small squared icon image in png/jpg format, size < 100x100px
 * `cover`: url to one or a list of cover images, typically a screenshot or a short gif which will show in the plugin store and the plugin installation dialog.

I think if we already have some images used elsewhere, `icon` and `cover` are recommended since they will show up in the plugin menu and the installation dialog.

thewtex:
> Looks great!
> 
> I added some suggestion for the docs to clarify some additional options and add basic examples. It would be also great if we can add for executable examples, e.g. binder, colab notebooks (like your itkwidgets notebooks), but that can be addressed in a following PR. Furthermore, we can think about how to best support other data types (xarray, ome-zarr etc) with optional compression.

:+1: thanks for the review, Wei! The itkwidgets binder, colab notebooks, xarray, ome-zarr support are coming next :fire: :-)

doc/content/docs/imjoy.md
```diff
@@ -0,0 +1,28 @@
+title: ImJoy Plugin
+---
+
+[![launch ImJoy](https://imjoy.io/static/badge/launch-imjoy-badge.svg)](http://imjoy.io/#/app?plugin=https://kitware.github.io/itk-vtk-viewer/app/)
+
+An *itk-vtk-viewer* plugin is available for [ImJoy](https://imjoy.io), a plugin powered hybrid computing platform for deploying deep learning applications such as advanced image analysis tools.
+
+## Installation
+
+Install the plugin in the [ImJoy Web App](http://imjoy.io/#/app?plugin=https://kitware.github.io/itk-vtk-viewer/app/) or [ImJoy Lite App](http://imjoy.io/#/lite?plugin=https://kitware.github.io/itk-vtk-viewer/app/) with the plugin URI:
```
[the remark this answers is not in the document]
thewtex: :+1: done.

doc/content/docs/imjoy.md
```diff
@@ -0,0 +1,28 @@
...
+
+```
+https://kitware.github.io/itk-vtk-viewer/app/
+```
+
+## Inputs
+
+Supported context `data` inputs:
+
+**image**: Image to be visualized. Can be an [itk.js Image](https://insightsoftwareconsortium.github.io/itk-js/api/Image.html) or a [scijs ndarray](http://scijs.net/packages/#scijs/ndarray).
```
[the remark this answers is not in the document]
thewtex: :+1: added.

doc/content/docs/imjoy.md
```diff
@@ -0,0 +1,28 @@
...
+
+## API functions
+
+In addition to the standard `setup` and `run` methods, the *itk-vtk-viewer* plugin provides the following methods:
+
+### setImage(image)
+
+Set the image to be visualized. Can be an [itk.js Image](https://insightsoftwareconsortium.github.io/itk-js/api/Image.html) or a [scijs ndarray](http://scijs.net/packages/#scijs/ndarray).
```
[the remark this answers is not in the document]
thewtex: :+1: added.

dist/index.html
thewtex: That is likely a corner use case, but, yes, the behavior may not be ideal. However, I am starting some major refactoring where we will address this.

dist/index.html
thewtex: Thanks for the tips! Added.

oeway:
Congrats! Also thanks for adding me to the author list.

Looks great!

It also works in ImageJ.JS (ImageJ compiled into JS plus ImJoy integration, a project I started recently): https://ij.imjoy.io/?plugin=https://kitware.github.io/itk-vtk-viewer/app/  

BTW, I am thinking about introducing plugin service discovery scheme for applications such as imagej.js. E.g. if we [register](https://imjoy.io/docs/#/api?id=apiregister) a `image-visualization` service from itk-vtk-viewer to the imjoy-core, plugins such as ImageJ.JS can then get all the imjoy plugins that supports `image-visualization`, and render them (including itk-vtk-viewer) in its plugin menu. Once clicked, ImageJ.JS converts `ImagePlus` object into a itkImage and visualize the image with ImJoy plugins such as itk-vtk-viewer. How does that sounds to you?

thewtex:
> BTW, I am thinking about introducing plugin service discovery scheme for applications such as imagej.js. E.g. if we register a image-visualization service from itk-vtk-viewer to the imjoy-core, plugins such as ImageJ.JS can then get all the imjoy plugins that supports image-visualization, and render them (including itk-vtk-viewer) in its plugin menu. Once clicked, ImageJ.JS converts ImagePlus object into a itkImage and visualize the image with ImJoy plugins such as itk-vtk-viewer. How does that sounds to you?

Sounds awesome!~! :rocket: :exploding_head:

PR #343, opened 2020-10-16
Multi scale

thewtex:


oeway:
This is awesome! Looking forward to it.

oeway:
Just for curiosity, can we feed a zarr store interface with `getItem` into the viewer. With imjoy-rpc, the zarr store interface can basically implemented in Python, just like zarr.js and vizarr does [example](https://colab.research.google.com/github/hms-dbmi/vizarr/blob/master/example/mandelbrot.ipynb)?

thewtex:
> Just for curiosity, can we feed a zarr store interface with getItem into the viewer. With imjoy-rpc, the zarr store interface can basically implemented in Python, just like zarr.js and vizarr does example?

@oeway to have access to the chunking interface, we will be able to provide a zarr interface that has a JavaScript implementation or a zarr interface that talks to an ImJoy-Python plugin. This is for performance.

thewtex:
@oeway a lot of refactoring in this branch, looking ahead to better kaibu integration...

Also fleshing out the ImJoy-Plugin API (now exported from the bundle). I would appreciate your feedback after it is ready.

oeway:
Wow, great features in the making! You even added layers support, cool!

thewtex:
@oeway yes, great suggestions -- getting around to them :-)

thewtex:
@oeway thank you for your patience! There are a number of fundamental improvements here. Of course, this gigantic branch is difficult to review. There is more to do -- looking forward to your feedback! I will follow up on issues.

PR #370, opened 2021-02-16
fix(applyRenderedImage): Create a copy of the color range before sending

thewtex:
Otherwise an Array reference is sent and all components end up with the
range of the last component.


oeway (approved):
Looks good!

PR #371, opened 2021-02-16
fix(independentComponents): Require explicit RGB or RGBA pixelType

thewtex:
Do not make an implicit assumption so independentComponents can be
enabled via pixelType.

### DIFF
--- a/src/UI/Layers/createLayersUIMachine.js
+++ b/src/UI/Layers/createLayersUIMachine.js
@@ -137,14 +137,10 @@ const assignImageContext = assign({
     if (actorContext.independentComponents === null) {
       // If a 2D RGB image
       if (
-        (context.use2D &&
-          componentType === IntTypes.UInt8 &&
-          (components === 3 || components === 4)) ||
         image.imageType.pixelType === PixelTypes.RGB ||
-          image.imageType.pixelType === PixelTypes.RGBA
+        image.imageType.pixelType === PixelTypes.RGBA
       ) {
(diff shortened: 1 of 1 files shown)


oeway:
When passing ndarray from the ImJoy interface, since there is no way to explicitly set the pixelType, maybe we should support additional options for the `setImage` function?

thewtex:
> When passing ndarray from the ImJoy interface, since there is no way to explicitly set the pixelType, maybe we should support additional options for the setImage function?

To prevent option creep/explosion on the viewer interface, can the object be converted, then the `pixelType` set?:

https://github.com/Kitware/itk-vtk-viewer/blob/94807c6385f1a4aec5d1e26d210292b56df13a62/src/utils.js#L11

oeway:
Hi, I don't fully get, how would you do it?

thewtex:
E.g.

```
itkImage = ndarrayToItkImage(ndarray)
itkImage.imageType.pixelType = 14
viewer.setImage(itkImage)
```

oeway:
Aha, well, yes and no, for js plugins we can potentially do that, but it will be hard for Python plugins where we just want to call `viewer.setImage(numpy_array)`. We may not want to always import itk module, or have the ndarray2ItkImage() function defined in the plugin.

thewtex:
Let's keep pushing forward Python plugins that provide the desired functionality.

Eventually in the config object, we will support an entire image configuration (data + complete config).

PR #373, opened 2021-02-16
fix(WebPack): Produce UMD library target

oeway:
With this PR we will be able to use the produce node module like this:
```js
import * as itkVtkViewer from "itk-vtk-viewer";
```

I am not sure whether this is the best way to do it, but I generate umd module for my other projects.


thewtex (approved):
@oeway thanks!!

PR #376, opened 2021-02-16
webpack

thewtex:
- chore(Dependencies): Bump webpack to 5.22.0
- fix(Webpack): Use UMD module for itkVtkViewerCDN.js


thewtex:
@oeway I think the previous change entered the unintended configuration.

oeway:
A note here, after this, we need to do `import * as itkVtkViewer from "itk-vtk-viewer/dist/itkVtkViewerCDN";` in order to use the UMD module.

PR #381, opened 2021-02-23
fix(applyRenderedImage): Default range for dependent components, no range

### LINKED ISSUE #380: Rendering bug for RGB images with only one channel
I found that if some of the channels are set to all 0, the viewer can easily go wrong:
For example this image: [_af647Cy5](https://user-images.githubusercontent.com/478667/108582412-76e24a00-7333-11eb-8dfb-6a97f7ed09c4.png)

If you download it and load it with the viewer https://kitware.github.io/itk-vtk-viewer/app/ it becomes all green, and complains an error: `itkVtkViewer.js:2 attempt to set zero width color range`.

Another observation is that, if we keep R channel, but set G and B to all 0, then it seems worked. If we keep only B or G, then the viewer does not work as expected.

@thewtex

thewtex:
All channels should have the same range for dependent components. Avoid
divide by 0 when the min/max range are the same.

Closes #380


oeway (approved):
Looks good, Thank you!

PR #384, opened 2021-02-25
fix(ImageRenderingActor): Multiscale rendering constraint on number of pixels

thewtex:
Have a reasonable limit for rendering.


thewtex:
@jourdain @bnmajor

PR #389, opened 2021-03-01
fix(package.json): fix commitizen config path to allow working with globally installed package

phcerdan:
Following instructions from https://github.com/commitizen/cz-cli


thewtex (approved):
:+1: gracias!

PR #391, opened 2021-03-01
feat(toMultiscaleChunkedImage): Support passing a Zarr store

thewtex:


oeway:
Hi @thewtex This is awesome, I am testing it now.

I noticed that you have these lines in the `extractScaleInfo` function:
```js
const zattrs = await store.getItem('.zattrs')
const multiscales = zattrs.multiscales
const name = multiscales[0]['name']
const datasets = multiscales[0].datasets
```

Are you assuming the `getItem` will return an json object directly?
Shouldn't it always be an ArrayBuffer? I think that will simplify the implementation of store key-agnostic. For example, this is a reference store implementation: https://gist.github.com/oeway/d8b5bad5cbb7bc234a87d2e8948d9164#file-vizarrreferencestore-imjoy-html-L49-L70

thewtex:
> Are you assuming the getItem will return an json object directly?

Yes, a json object or an arraybuffer.

> Shouldn't it always be an ArrayBuffer? 

This makes things more complex and adds unnecessary conversions, correct? Why force JSON data into an ArrayBuffer? And then convert it back? With Zarr, we know that some keys are always JSON and some are always ArrayBuffer's.

thewtex:
While it seems to add unnecessary overhead, I'll change the behavior to be consistent with the Zarr Python spec and behavior, which is always using `bytes` for values.

oeway:
Yes, I agree that will be some overhead, but perhaps that doesn't matter that much because those are mostly for meta information which often very small.

I was about to say the same, since the Zarr store in Python also return bytes.

thewtex:
@oeway thanks for the review and suggestion. All ArrayBuffer values is now implemented. Please take a look.

oeway:
Hey, thanks a lot! Super exciting to try it! 

The PR looks good! I just trie to test it but don't have the right data format yet.

thewtex:
Thanks! With the foundation down, I'll continue with the other store implementations :-)

PR #393, opened 2021-03-02
build(Webpack): Enable devServer writeToDisk

thewtex:
This should help with the ImJoy tests when `npm run build` is not run
before `npm test`.


thewtex:
@phcerdan

PR #397, opened 2021-03-05
Add cropping planes controlled by 3D widget or viewer API

thewtex:
Volumes, image slices, geometries and point sets are cropped (aka clipped) by a shared box of planes.  Adding a new image/geometry will reset the crop box to encompass all data in the scene.  

Can set 6 arbitrary clip planes with a function on the viewer object.
```
const [iMin, iMax, jMin, jMax, kMin, kMax] = [0, 3, 0, 3, 0, 3]
const origin = [iMin, jMin, kMin]
// opposite corner from origin
const corner = [iMax, jMax, kMax]
const planes = [
  // X min/max
  { normal: [1, 0, 0], origin },
  {
    normal: [-1, 0, 0],
    origin: corner,
  },
  // Y min/max
  { normal: [0, 1, 0], origin },
  {
    normal: [0, -1, 0],
    origin: corner,
  },
  // X min/max
  { normal: [0, 0, 1], origin },
  {
    normal: [0, 0, -1],
    origin: corner,
  },
]
viewer.setCroppingPlanes(planes)
```


thewtex (commented):
:+1:

thewtex:
@PaulHax thank you!!

Looks good!

History is complicated -- we could forgo a squash in the interest of getting it in.

Only item I noticed in review that debug builds of the wasm modules crept in. I will update to the latest itk-wasm (includes new optimizations), and push Release builds.

Issue #407, opened 2021-06-23
Support displaying pointSets


oeway:
Hi @thewtex I am trying to passing a point cloud to the viewer through the ImJoy API, e.g. I have a Float32Array with xyz coordinates, how can I feed that into `crreateViewer` as `pointSets`? A code snippet would be great! Thanks in advance.

PR #408, opened 2021-06-24
feat(ImJoy): Support setPointSets

oeway:
This PR allows displaying point cloud with the ImJoy API.

Fixes https://github.com/Kitware/itk-vtk-viewer/issues/407


oeway:
@thewtex One of the tests is failing and I have no idea how to fix it. Please take a look.

thewtex:
@oeway awesome work!! 🌟 ☁️ 🌻 

>@thewtex One of the tests is failing and I have no idea how to fix it. Please take a look.

We just have to bump the expected number of tests from `5` to `6` here:

https://github.com/Kitware/itk-vtk-viewer/blob/718ad2b482f2a30ee5556b24f45cc5c18b7a8348/test/imjoyTest.js#L53

>Another question is about the Plane Index Slider, how do we disable that for pointSets?

In progress, but the planes will be used to clip the point sets so you can interactively "look" into them.

oeway:
Thank you!

I see! I tried the plane index slider, it doesn’t seem to have any effect.
In any case it might be good to provide a function to disable it.

oeway:
@thewtex I would like to display several pint sets, and I am wondering how we can specify the color for each set.

PR #456, opened 2022-04-01
fix(ndarraytoitkimage.js): directly set direction as Float32 arrays

bnmajor:
`itk-wasm` expects that direction have a float array set directly on the key rather than the old approach of setting the array within a data object.


thewtex (approved):
:+1:

PR #470, opened 2022-04-21
dandi zarr

thewtex:
- fix(createViewerFromUrl): url undefined for image, labelImage
- fix(isZarr): Support DANDI urls


PaulHax (approved):
LGTM.   Good catch on `createViewerFromUrl`

PR #472, opened 2022-04-21
fix(ngff zarr): dot is default dim separator, cap voxel count

PaulHax:
If no dimension_separator in .zarray, assume '.'

Limit loading next multi scale level if voxel count over
512 * 512 * 512


thewtex (approved):
AAAAwesome! Thank you @PaulHax !

PR #474, opened 2022-04-22
fix(SlicingPlanes): Enable by default

thewtex:
Enable slicing planes by default in the 3D view.

The slicing planes are a helpful visualization. And, until we refine our
default gradient opacity setting for all datasets, there will always be
something other than a blank scene (if it is too high).

BREAKING_CHANGE: 3D slicing planes are enabled by default


PaulHax (commented):
Good idea.  Boo test failure.

PR #475, opened 2022-04-25
build(webpack): set the publicPath to an empty string

bnmajor:
Automatic publicPath is not always supported.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -73,6 +73,7 @@ module.exports = (env, argv) => [
     devtool: argv.mode === 'development' ? 'eval-source-map' : 'source-map',
     output: {
       filename: 'itkVtkViewer.js',
+      publicPath: '',
     },
     resolve: {
       alias: {


PaulHax (commented):
A step in the right direction!  

Regarding the test 404: Looks like the default, `publicPath: 'auto'` resolves to an absolute path which helps out pipeline.worker.js

bnmajor:
@PaulHax Thanks! If you have some insight here I'd love it since my experience with Webpack is limited and I'm basically going through a lot of Google and guess and check at this point...

I need to add `itk-vtk-viewer` as a dependency for [tensorboard-plugin-3d](https://github.com/KitwareMedical/tensorboard-plugin-3d), but when I do I get an error: `Automatic publicPath is not supported in this browser`. Local testing shows that setting the `publicPath` to an empty string (generally the first suggestion I am seeing when digging around) does work for me in the plugin, but this is creating test failures (as you can see).

Running the tests myself I see that it fails in `imjoyTest.js` trying to call `setImage` with a URL. When the publicPath is not set to `auto` it returns a 404 trying to fetch `NrrdImageIOReadImage.wasm`. It should be using the path `/base/dist/itk/image-io/NrrdImageIOReadImage.wasm` but instead it is looking for `/base/dist/itk/web-workers/min-bundles/itk/image-io/NrrdImageIOReadImage.wasm`.

PaulHax:
I'm been flaying at this same thing too =(

Not sure this is the proper approach but lifting pipeline.worker.js up passes the tests =)

https://github.com/Kitware/itk-vtk-viewer/compare/master...PaulHax:public-path-hax?expand=1

bnmajor:
This PR is outdated and not needed, closing.

PR #477, opened 2022-04-26
feat(url parameters): add gradientOpacity url param option

PaulHax:
A test:
http://localhost:8082/?image=https://dandiarchive.s3.amazonaws.com/zarr/3d313fc2-0204-496d-bfa1-5c90951ee640&gradientOpacity=.3

### DIFF
--- a/src/IO/processFiles.js
+++ b/src/IO/processFiles.js
@@ -36,20 +36,9 @@ import vtkITKHelper from 'vtk.js/Sources/Common/DataModel/ITKHelper'
 import UserInterface from '../UserInterface'
 import createViewer from '../createViewer'
 
-function typedArrayForBuffer(typedArrayType, buffer) {
-  let typedArrayFunction = null
-  if (typeof window !== 'undefined') {
-    // browser
-    typedArrayFunction = window[typedArrayType]
-  } else {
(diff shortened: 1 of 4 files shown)


src/IO/processFiles.js
```diff
@@ -60,27 +49,29 @@ export const processFiles = async (
...
 }) => {
   let readDICOMSeries = readImageDICOMFileSeries
-  if (files.length < 2 || !!!image) {
+  if (files.length < 2 || !image) {
     readDICOMSeries = function() {
       return Promise.reject('Skip DICOM series read attempt')
     }
   }
   try {
-    const { image: itkImage, webWorkerPool } = await readDICOMSeries(files)
```
thewtex, line 83: I think we need to keep `webWorkerPool` around, also call `webWorkerPool.terminate()`, correct?

thewtex (changes requested):
@PaulHax awesome!! :100: thank you!

One possible issue noted inline.

src/IO/processFiles.js
PaulHax: Ah yes.  Good catch.

thewtex (approved):

PR #480, opened 2022-05-04
feat(ngff zarr): add multi channel support

PaulHax:
Somehow these hacks work for all the datasets I've loaded except for the 2 lowest resolution scales in oeway's astronaut where the chunks span components.


thewtex:
> const chunkSize = toDimensionArray(['c', 'x', 'y', 'z'], info.chunkSize)

Instead of `['c', 'x', 'y', 'z']`, this should be `info.dims` now, correct?

PaulHax:
Works on oeway's astronaut now.  Some fishy stuff with the x chunk index to shift down x rows:
```
pixelArray[
                cc +
                  zz * pixelStrides.z +
                  (yy + x) * pixelStrides.y +
                  xx * pixelStrides.x
              ] =
```

oeway:
Thanks for working on this! Looking forward it! FYI: I am currently making a [browser file system](https://fm.imjoy.io/) that can handle large files backed by indexeddb. Would be nice to test using pyodide/jupyterlite + zarr-python to read in-browser large files, then pipe the data as a zarr store to itk-vtk-viewer!

src/IO/ImageDataFromChunks.worker.js
```diff
@@ -46,74 +42,75 @@ registerWebworker().operation(
...
 
+    const arrayShape = Object.fromEntries(info.arrayShape)
+    const pixelStrides = {
+      z: arrayShape.c * arrayShape.x * arrayShape.y,
+      y: arrayShape.c * arrayShape.x,
+      x: arrayShape.c,
+    }
+
+    const chunkSize = Object.fromEntries(info.chunkSize)
+    const chunkStrides = {
```
thewtex, line 53: The chunk strides for each dimension will be based on the input MultiscaleSpatialImage `dims` and their order: maybe a `slice` here on the `dims` array starting on the `dim` index and a `reduce` based on the `chunkSize`.

src/IO/ImageDataFromChunks.worker.js
```diff
@@ -46,74 +42,82 @@ registerWebworker().operation(
...
+                  (xx - x * chunkSize.x) * chunkStrides.x +
+                    (yy - y * chunkSize.y) * chunkStrides.y +
+                    (zz - z * chunkSize.z) * chunkStrides.z +
+                    (cc - c * chunkSize.c) * chunkStrides.c
+                ]
+            } // column
+          } // row
+        } // slice
+      } // component
+    } // chunk
```
thewtex, line 120: Optimizations to add to this critical piece:

[ ] 1. Three versions of the loop based on the dim order. a) tczyx, b) tzyxc c) other. a) for most OME-NGFF, b for the cancel contiguous, we may encounter c) in the wild, e.g. it is also common to have tzcyz. c) can be like it is now. a) and b) will further be optimized in the following steps.

[ ] 2. For the conditions where we can use them in the inner-most loop, use `.subarray` and `.set` for getting and setting the contiguous region.

[ ] 3. Cache the offsets in the intermediate loops.

thewtex (changes requested):
:t-rex: :taco: :tada:

Looks awesome!

Steps for optimization noted inline that I think we should add before merging.

src/IO/ImageDataFromChunks.worker.js
PaulHax: Added caching of offsets.  Do we have any example images of non `tczyx` dim ordering we would like to support?  I have not tested any...

thewtex:
Looking good. Can we base and add tests for the new datasets?

PaulHax:
> Looking good. Can we base and add tests for the new datasets?

Added tests for a selection of the working images.

Some issues:
- All the first_instar_brains: .zattrs 'axes' metadata order does not match the array's .zattrs "_ARRAY_DIMENSIONS".  I'll change code to use _ARRAY_DIMENSIONS.  
- first_instar_brain_czyx.zarr/.zattrs file is empty {}
- Most of the idr datasets have their dtype endianness the way we don't support, `>u2` and `>f4`.  Not sure what to do there.

thewtex:
For the first two, we should fix the data generator. Could you please create issues on the multi-scale-spatial-image repository, and assign it to me?

For the endianness issue, we should transform an chunk array from big to little before loading into the itk.Image. Could you please look for a JS library to do that and apply when required?

src/IO/dtypeUtils.js
```diff
@@ -0,0 +1,81 @@
...
+  ['i1', 'getInt8'],
+  ['u2', 'getUint16'],
+  ['i2', 'getInt16'],
+  ['u4', 'getUint32'],
+  ['i4', 'getInt32'],
+  ['f4', 'getFloat32'],
+  ['f8', 'getFloat64'],
+])
+
+export const ElementGetter = (dtype, buffer) => {
```
thewtex, line 73: For efficiency, let's have this operate over the chunk TypedArray instead of the element

thewtex (changes requested):
Looking good.

PaulHax:
> [ ] 2. For the conditions where we can use them in the inner-most loop, use .subarray and .set for getting and setting the contiguous region.

Implemented the above with guidance from the old code.  Good speedup with `first_instar_brain_zyxc.zarr` highest resolution taking the worker ~11ms vs old way of ~26ms.

src/IO/ZarrMultiscaleSpatialImage.js
```diff
@@ -87,13 +87,11 @@ const makeCoords = ({ shape, multiscaleImage, dataset }) => {
 const createScaledImageInfo = ({
   multiscaleImage,
   pixelArrayMetadata,
-  pixelArrayAttrs,
   dataset,
 }) => {
   const dims =
-    pixelArrayAttrs?._ARRAY_DIMENSIONS ?? // xarray
```
thewtex: Should we just change the order here -- first try / trust `axes`, then `_ARRAY_DIMENSIONS`?

thewtex (approved):
Turbo-tacular!~  :rocket: :man_singer: 

Please take a look at a comment inline, follow-up as needed, and merge this puppy. :pleading_face:

src/IO/ZarrMultiscaleSpatialImage.js
PaulHax: Seems that many NGFFs don't have a foo.zarr/s0/.zattrs where `_ARRAY_DIMENSIONS` lives.  To avoid 404s the code is not fetching the scale level `.zattrs`, so no `_ARRAY_DIMENSIONS` to fall back to.

oeway:
Hi @PaulHax @thewtex Awesome to see this PR get merged! Well done, can't wait to try it.

It looks like the master branch CI is failing and that it explains why I cannot already show the multi-channel zarr images.

thewtex:
Hi @oeway  CI was given some :heart: , v12.3.0 is now out with these improvements, please give it a whirl!

PR #486, opened 2022-05-17
ci(Data): Connect to pinata.cloud hosts

thewtex:
Testing data has been pinned there too for reliability and speed.


PaulHax (approved):
caliente!

PR #489, opened 2022-05-19
feat(Zarr): Identify RGB RGBA pixel types

thewtex:
2D, uint8 pixel and 3 or 4 components -> Assume we want to render like
an RGB / RGBA.


thewtex:
@PaulHax I did not test this -- are any of the testing dataset we have relevant?

PaulHax:
The removed logo.zarr matched that signature.

PR #491, opened 2022-05-23
itkwidgets type support

thewtex:
- feat(ndarrayToItkImage): Use itk-wasm types, add 64bit int support
- doc(ImJoy): itk.js -> itk-wasm
- feat(imJoy): Add Codec for decoding itk-wasm Image from itkwidgets


thewtex:
@oeway

PR #499, opened 2022-06-01
Brighten 2D slices and upgrade vtk.js from 23.x to 24.x

PaulHax:
To increase the brightness of 2D slices with multiple components,  add component colors to final result without multiplying by a normalized weight. 

Also sets a flat initial Gaussian for 2D, rather than inverted, to avoid blowing out non-independent style images.

Setting setUseLookupTableScalarRange(true) on slices for component specific window/level.

BREAKING CHANGE: 2D slice independent component mixing approach changed. Slice images will be brighter as components add to the final color without being normalized by component weight.


thewtex (commented):
@PaulHax nice work!!

Yes, a custom mapper makes sense.

Can you please remove Node v14 from CI, which is causing failures?

PaulHax:
> Can you please remove Node v14 from CI, which is causing failures?

I removed Node v14.

thewtex (approved):
:1st_place_medal:

PR #504, opened 2022-06-14
Fuse multiple images into Conglomerate image components

PaulHax:
To view multiple overlapping images at once: load an array of input images and "fuse" their components into a single vtkImageData in updateRenderedImage().  Can pass an array of images to `createViewer`
```
itkVtkViewer.createViewer({
   images: [imageUrlA, imageUrlB]
})
```
Or give comma separated list of images addresses to the url parameter
`
http://localhost:8082/?image=test-data/HeadMRVolume.nrrd,test-data/HeadMRVolumeLabels.nrrd
`

Also, to support 4 components, up from 3, stop ImageMapper from generating mip-maps.

BREAKING CHANGE: readFiles() no longer returns { image: MultiscaleSpatialImage, ... }
Instead returns { images: MultiscaleSpatialImage[], ... }. Also readFiles() no longer reads the `image` paramter, it reads `images`  like so `readFiles({ images: [imageUrlA, imageUrlB] })`


oeway:
Maybe related to this, do you think we can have an ImJoy api to allow add image to the viewer?

PaulHax:
> Maybe related to this, do you think we can have an ImJoy api to allow add image to the viewer?

Looks like `setImage()` is wired up for ImJoy.  Kind of separate thing if that has issues.

src/IO/MultiscaleSpatialImage.js
```diff
@@ -5,6 +5,9 @@ import componentTypeToTypedArray from './componentTypeToTypedArray'
 import WebworkerPromise from 'webworker-promise'
 import ImageDataFromChunksWorker from './ImageDataFromChunks.worker'
 import { CXYZT, toDimensionArray } from './dimensionUtils'
+
+export const MAX_COMPONENT_COUNT = 4
```
thewtex, line 9: We have the limitation in vtk.js rendering for a max component count from WebGL, but this is imposed with the number of selected components that are enabled on the input image. Those selected components are used to generate the fusedImage, which results in an in-memory vtkImageData with a potential subset of components. Can we keep that functionality? That is, ConglomerateMultiscaleSpatialImage can have N components. A MAX_COMPONENT_COUNT is a limitation of the vtk.js renderer, and the selected component subset is defined in the rendering and rendering options.

PaulHax: OK, I'll fix.

thewtex: :pray:

thewtex (approved):

thewtex:
> Interesting test case with 27 components:

Add to test suite?

thewtex (changes requested):
Todos:

- [ ] Do the conglomerate generation between url parsing and `createViewer`
- [ ] Revert the `readFiles` to previous API
- [ ] Add the conglobmerate generation functionality to the export bundle
- [ ] Add test with the many datasets (#507 )

PR #520, opened 2022-07-18
Emit a signal with the screenshot png when one is generated

bnmajor:
Needed for PR InsightSoftwareConsortium/itkwidgets#448, issue InsightSoftwareConsortium/itkwidgets#445

### DIFF
--- a/src/createViewer.js
+++ b/src/createViewer.js
@@ -180,6 +180,9 @@ const createViewer = async (
           case 'Z_SLICE_CHANGED':
             eventEmitter.emit('zSliceChanged', event.data)
             break
+          case 'SCREENSHOT_TAKEN':
+            eventEmitter.emit('screenshotTaken', event.data)
+            break
           default:
             throw new Error(`Unexpected event type: ${event.type}`)
         }
@@ -656,6 +659,7 @@ const createViewer = async (
     'pointSetOpacityChanged',
     'pointSetSizeChanged',
     'pointSetRepresentationChanged',
+    'screenshotTaken',
   ]
 
   publicAPI.getEventNames = () => eventNames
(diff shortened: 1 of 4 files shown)


src/createViewer.js
```diff
@@ -177,6 +177,9 @@ const createViewer = async (
           case 'Z_SLICE_CHANGED':
             eventEmitter.emit('zSliceChanged', event.data)
             break
+          case 'TAKE_SCREENSHOT':
+            eventEmitter.emit('screenshotTaken', publicAPI.captureImage())
```
thewtex, line 181: To use the higher quality screenshot generated in *takeScreenshot.js*, can we emit the base64 PNG taken there in the `data` payload of a `SCREENSHOT_TAKEN` event, then use that here?

thewtex (changes requested):

src/createViewer.js
bnmajor: That is what this is currently emitting. The selection is [listened for](https://github.com/InsightSoftwareConsortium/itkwidgets/blob/e09abe7d451bec19d2ea5baf2aecceadd09116b5/itkwidgets/viewer.py#L67-L69) and then base64 PNG is passed to the [update function](https://github.com/InsightSoftwareConsortium/itkwidgets/blob/e09abe7d451bec19d2ea5baf2aecceadd09116b5/itkwidgets/viewer.py#L81-L93). Unless I am misunderstanding what you are asking for?

thewtex: > That is what this is currently emitting. The selection is listened for and then base64 PNG is passed to the update function. Unless I am misunderstanding what you are asking for?

These look good -- I am referencing storing the output here:

https://github.com/Kitware/itk-vtk-viewer/blob/c83039cd7eb65ec8663a3f3cc68d7db0c9c5e996/src/Rendering/VTKJS/Main/takeScreenshot.js#L11

which will benefit from:

https://github.com/Kitware/itk-vtk-viewer/blob/c83039cd7eb65ec8663a3f3cc68d7db0c9c5e996/src/Rendering/VTKJS/Main/takeScreenshot.js#L6-L9

Get the output, then emit as a signal as `service.send({ type: 'SCREENSHOT_TAKEN', data: base64PNG })`

thewtex (approved):

PR #522, opened 2022-07-20
Set label image

bnmajor:
Add `setLabelImage` and `getLabelImage` functions to both the viewer and ImJoy plugin API.

### DIFF
--- a/src/createViewer.js
+++ b/src/createViewer.js
@@ -1052,6 +1052,26 @@ const createViewer = async (
     return actorContext.colorMaps.get(componentIndex)
   }
 
+  publicAPI.setLabelImage = async labelImage => {
+    const imageName = context.images.selectedName
+    const multiscaleLabelImage = await toMultiscaleSpatialImage(
+      labelImage,
+      true
+    )
+    if (multiscaleLabelImage.name === 'Image') {
+      multiscaleLabelImage.name = 'LabelImage'
+    }
+    service.send({
+      type: 'ADD_LABEL_IMAGE',
+      data: { imageName, labelImage: multiscaleLabelImage },
+    })
+  }
+
+  publicAPI.getLabelImage = () => {
+    const name = context.images.selectedName
+    return context.images.actorContext.get(name).labelImage
+  }
+
   publicAPI.setLabelImageLookupTable = (lookupTable, name) => {
     if (typeof name === 'undefined') {
       name = context.images.selectedName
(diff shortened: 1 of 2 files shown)


thewtex (approved):
:+1:

PaulHax:
@bnmajor 

I did not reproduce the long load time issue.  This is my test:
```
<!DOCTYPE html>
<html>
  <head>
    <meta http-equiv="Content-type" content="text/html; charset=utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
  </head>
  <body>
    <div
      style="position: relative; border: 1px solid red; width: 1000px; height: 1000px;"
      id="viewport"
    ></div>

    <script type="text/javascript" src="itkVtkViewer.js"></script>
    <script>
      async function main() {
        const imageUrl = new URL(
          'test-data/HeadMRVolume.nrrd',
          window.location.origin
        )
        const labelUrl = new URL(
          'test-data/HeadMRVolumeLabels.nrrd',
          window.location.origin
        )
        const container = document.querySelector('#viewport')
        const viewer = await itkVtkViewer.createViewer(container, {
          image: imageUrl,
          rotate: false,
        })
        viewer.setLabelImage(labelUrl)
      }

      main()
    </script>
  </body>
</html>
```
Maybe I should try a specific dataset?

bnmajor:
@PaulHax Yes, there seems to be one dataset in particular that is having this issue. I'll share the Google Drive link with you

PaulHax:
Looks like the label map image flow never gets past the label map downsample workers here: https://github.com/Kitware/itk-vtk-viewer/blob/master/src/IO/InMemoryMultiscaleSpatialImage.js#L280

Probably cuz the label map image in question a Float64 array, not Int

thewtex:
We should have Float64 support:

https://github.com/Kitware/itk-vtk-viewer/blob/c83039cd7eb65ec8663a3f3cc68d7db0c9c5e996/src/IO/Downsample/DownsampleLabelImage.cxx#L180

What happens if the image is cast locally to an int pixel type?

PaulHax:
> What happens if the image is cast locally to an int pixel type?

Same behavior (downsample take a long time) If I pass the label image through this:
```js
const newArray = new Int32Array(image.data.length)
for (let i = 0; i < image.data.length; i++) {
  newArray[i] = image.data[i]
}
image.data = newArray
image.imageType.componentType = IntTypes.Int32 //'int32'
```

thewtex:
I will look into the poor performance on that specific dataset, tracked in #527

thewtex (approved):

PR #529, opened 2022-07-22
Fixes for fuseComponents and component checkbox UI

PaulHax:
When 4 components were ON and user clicked checkbox of OFF component,
sometimes it turned on, sometimes it did not.  Fixed.

If new fused image takes some time to build in worker and user changes
rendering parameters of current image, there would be an error as the
current image buffer was transferred to worker.  Changed to no longer transfer array.


thewtex (approved):
:+1: :100:

PR #540, opened 2022-08-06
Label downsample

thewtex:


PaulHax (commented):
itk-wasm is nifty!  

Thanks for fixing the labelImage only bug.  Can we incorporate this PR with fix in another area of the code and a test?
https://github.com/thewtex/itk-vtk-viewer/pull/1

bnmajor (approved):
:tada: Awesome!

PR #541, opened 2022-08-08
Load subset of image scale defined by crop planes

PaulHax:
Can narrow the crop planes and the viewer will load a subset of the image if the image scale changes.

Converts MultiscaleSpatialImage::scaleLargestImage(scale) to getImage(scale, worldBounds).
getImage() converts world bounds to image index bounds, fetches appropriate chunks,
returns ITK image fitting inside of world bounds.  If previously cached scale image
contains requested bounds, getImage returns larger cached image.

Harmonizes return value of InMemoryMultiscaleSpatialImage::getChunks
with ZarrMultiscaleSpatialImage::getChunks.  But InMemoryMultiscaleSpatialImage::getChunks
is never called as whole image for each scale is cached in constructor.

Changing the crop planes triggers adjustScaleForFramerate ImageRenderingActor
state if the user has not manually selected a scale.


src/internalUtils.js
```diff
@@ -0,0 +1,15 @@
+import { vec3 } from 'gl-matrix'
+import { computeBoundsFromPoints } from 'vtk.js/Sources/Common/Core/Math'
+import vtkBoundingBox from 'vtk.js/Sources/Common/DataModel/BoundingBox'
+
+export const transformBounds = (transformingMat4, bounds) => {
```
thewtex, line 5: To keep semantically meaningful modules / file names, please put this into a `transformBounds` file / module.

thewtex (changes requested):
Awesome @PaulHax ! :star2: 

A few minor issues inline.

PaulHax:
> A few minor issues inline.

Addressed!

thewtex (approved):

thewtex (approved):

PR #542, opened 2022-08-11
ci(Data): Use tarball for CI fetch

thewtex:
The IPFS node has difficulty downloading the data without a timeout.
This may be due to starting up the node and immediately requesting large
amounts of data without the opportunity to provide data to peers.

Download and extract a versioned tarball for now from GitHub Releases.

In the future, we will look to download a versioned, provenance rich
DataLad dataset.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -20,33 +20,14 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Download testing data
+        shell: bash
+        run: |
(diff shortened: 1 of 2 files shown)


thewtex:
@PaulHax @bnmajor @labdala a temporary workaround to get our CI clean / deployments published.

Additional improvements coming re:

https://github.com/InsightSoftwareConsortium/itk-viewer-testing-data
https://github.com/datalad/datalad/issues/6929

PR #546, opened 2022-08-22
fix(croppingPlanes): disable picking on volume actor

PaulHax:
Allows for grabbing of crop plane handles "through" the volume bounding box.  

Sad effect is can click on handles even if they are totally occluded by the volume!

Alternatives:
- change appearance of handles when on other side of volume
- fix vtk.js picking (click through volume opacity threshold?)


thewtex (approved):
:ghost:

PR #547, opened 2022-08-23
fix(createImageRenderingActor): handle errors making new image scale

### LINKED ISSUE #544: Check for request of too large of an image

PaulHax:
Update scale drop-down to last successfully loaded scale after error loading user selected scale.

Good test case:
http://localhost:8082/?rotate=false&image=https://dandiarchive.s3.amazonaws.com/zarr/3d313fc2-0204-496d-bfa1-5c90951ee640
Then select scale 3 for "Out of memory error transferring to worker" or scale 2 for "to many network requests" error.

closes #544

### DIFF
--- a/src/Context/ImageActorContext.js
+++ b/src/Context/ImageActorContext.js
@@ -2,8 +2,11 @@ class ImageActorContext {
   // MultiscaleSpatialImage to be visualized
   image = null
 
-  // The rendered image / label image scale
+  // The target image scale
   renderedScale = null
+
+  // The successfully loaded scale
+  loadedScale = null
 
   // Automatically adjust the rendered scale
   isFramerateScalePickingOn = true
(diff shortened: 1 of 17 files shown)


src/Context/ImageActorContext.js
```diff
@@ -2,9 +2,12 @@ class ImageActorContext {
   // MultiscaleSpatialImage to be visualized
   image = null
 
-  // The rendered image / label image scale
+  // The target image scale
   renderedScale = null
```
thewtex, line 6: To be accurate with the naming here, should this be called `targetScale` and `loadedScale` -> `renderedScale` ?

PaulHax: Good idea.  Changed.  It breaks the Bootstrap and Material GUIs though 🤦

thewtex: Thanks -- could you please address Bootstrap and Material after merge?

PaulHax: Will do.

thewtex (approved):

PR #548, opened 2022-08-24
fix(MultiscaleSpatialImage): cache only one image per scale

### LINKED ISSUE #543: Limit image caching

PaulHax:
Before was keeping images that were not fully contained by new image.

closes #543


thewtex (approved):
:+1: 

We can get crazy later :woozy_face:

PR #555, opened 2022-08-26
Automatically pick finer or coarser scale when changing bounds

PaulHax:
Big change to the `ImageRenderingActor`

### DIFF
--- a/src/Rendering/Images/createImageRenderingActor.js
+++ b/src/Rendering/Images/createImageRenderingActor.js
@@ -20,7 +20,7 @@ const assignHigherScale = assign({
   images: context => {
     const images = context.images
     const actorContext = images.actorContext.get(images.updateRenderedName)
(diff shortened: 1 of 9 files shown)


src/Rendering/Images/createImageRenderingActor.js
```diff
@@ -16,91 +19,82 @@ const assignUpdateRenderedNameToSelectedName = assign({
   },
 })
 
-const assignHigherScale = assign({
-  images: context => {
-    const images = context.images
-    const actorContext = images.actorContext.get(images.updateRenderedName)
-    actorContext.renderedScale--
-    return images
+const assignFinerScale = assign({
```
thewtex, line 22: I like the finer / coarser verbage :+1: :book:

thewtex (approved):
:black_square_button: :microscope:

PR #557, opened 2022-09-02
Extract vtkjs from reference-ui

PaulHax:
Setting changing piecewise function points via API, UI TF editor, or applyColorRange triggers mapToPiecewiseFunctionNodes which changes the vtk.js opacity function object.
    
Event flow:
    [API | UI | applyColorRange] -> IMAGE_PIECEWISE_FUNCTION_POINTS_CHANGED
    -> mapToPiecewiseFunctionNodes -> IMAGE_PIECEWISE_FUNCTION_CHANGED


src/UI/Images/transferFunctionManipulators.js
```diff
@@ -0,0 +1,177 @@
+import vtkMouseRangeManipulator from '@kitware/vtk.js/Interaction/Manipulators/MouseRangeManipulator'
```
thewtex, line 1: Next steps: To UI / Renderer independent manipulator that works with DOM events and renderer viewport state, etc. stored in the context, and send events to the machine.

thewtex (approved):
:clap: :medal_sports:

PR #559, opened 2022-09-04
fix(ItkVtkViewProxy): grab crop handles when animating

PaulHax:
Add subclassed WidgetManager to handle events even when interactor is
animating (like with FPS speed checks or auto rotating).

The FPS speed check after each crop handle change triggered the renderWindow.interactor
to be animating which disabled the WidgetManager from handling events.


thewtex (approved):
:1st_place_medal: :star:

PR #560, opened 2022-09-05
fix(ImageRenderingActor): ignore stale updateRenderedImage results

PaulHax:
User set bounds and scale could get out of sync with rendering.  Steps:
1. Trigger slow updateRenderedImage by setting super fine scale.
2. Trigger fast updateRenderedImage by setting coarse or cached scale.
3. Fast updateRenderedImage sets rendering, then stale slow updateRenderedImage
overrides the more up to date updateRenderedImage result.

Update context and the rendering after updateRenderedImage Promise
resolves.  Promise result is ignored if loadingImage
state has been exited.


thewtex (approved):
:+1:

PR #563, opened 2022-09-12
Load image bounded by camera frustum

PaulHax:
For the world bounds passed to `MultiscaleSpatialImage.getImage()`, `updateRenderedImage()` intersects the world axis aligned bounding box of the camera frustum with the cropping planes.  Camera modified events trigger `updateRenderedImage()` after debounce.

Changes colorRange and colorRangeBounds to automatically grow after loading more image data.  As image scales become finer, intensity extremes emerge. If the colorRange/colorRangeBounds are changed by the UI or API, they are "touched" and don't automatically change as new data is loaded. Before was calculating the range of data just once on the coarsest scale.  However, that may have had a better result for visualization by clipping the extremes in the transfer functions.

### DIFF
--- a/src/Context/ImageActorContext.js
+++ b/src/Context/ImageActorContext.js
@@ -45,10 +45,14 @@ class ImageActorContext {
   // Map of image intensity component to array of [minValue, maxValue] for
   // mapping colors
   colorRanges = new Map()
+  // If not touched, keep growing ranges as new parts of the image are loaded
+  colorRangesTouched = new Map()
 
   // Map of image intensity component to array of [minBound, maxBound] for
   // limiting the color range in the UI
   colorRangeBounds = new Map()
+  // If not touched, keep growing bounds as new parts of the image are loaded
+  colorRangeBoundsTouched = new Map()
 
   // Map of the image intensity component to an object representing the
   // piecewise function. This object has two properties: range and nodes.
--- a/src/Context/ImageActorContext.js
+++ b/src/Context/ImageActorContext.js
@@ -2,9 +2,6 @@ class ImageActorContext {
   // MultiscaleSpatialImage to be visualized
   image = null
 
-  // The target image scale
-  targetScale = null
-
   // The successfully loaded scale
   loadedScale = null
 
@@ -45,14 +42,16 @@ class ImageActorContext {
   // Map of image intensity component to array of [minValue, maxValue] for
   // mapping colors
   colorRanges = new Map()
-  // If not touched, keep growing ranges as new parts of the image are loaded
-  colorRangesTouched = new Map()
+  // Keep growing component ranges as new parts of the image are loaded
+  // Map of image intensity component to Boolean
+  colorRangesAutoAdjust = null
 
   // Map of image intensity component to array of [minBound, maxBound] for
   // limiting the color range in the UI
   colorRangeBounds = new Map()
-  // If not touched, keep growing bounds as new parts of the image are loaded
-  colorRangeBoundsTouched = new Map()
+  // Keep growing component bounds as new parts of the image are loaded
+  // Map of image intensity component to Boolean
+  colorRangeBoundsAutoAdjust = null
 
   // Map of the image intensity component to an object representing the
   // piecewise function. This object has two properties: range and nodes.
--- a/src/IO/ConglomerateMultiscaleSpatialImage.js
+++ b/src/IO/ConglomerateMultiscaleSpatialImage.js
@@ -21,11 +21,18 @@ export class ConglomerateMultiscaleSpatialImage extends MultiscaleSpatialImage {
     this.images = images
   }
 
-  async buildImage(scale, bounds) {
-    // Run sequentially rather than Promise.all to avoid hang
+  async getImage(scale, worldBounds = []) {
+    this.worldBoundsStashedForOurBuildImage = worldBounds
+    return super.getImage(scale, worldBounds)
+  }
+
+  async buildImage(scale /*bounds*/) {
+    // Run sequentially rather than Promise.all to avoid hang during chunk fetching
     const builtImages = []
     for (const image of this.images) {
-      builtImages.push(await image.buildImage(scale, bounds))
+      builtImages.push(
+        await image.getImage(scale, this.worldBoundsStashedForOurBuildImage)
+      )
     }
     return builtImages
   }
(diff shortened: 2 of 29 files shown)


src/Context/ImageActorContext.js
```diff
@@ -45,10 +45,14 @@ class ImageActorContext {
   // Map of image intensity component to array of [minValue, maxValue] for
   // mapping colors
   colorRanges = new Map()
+  // If not touched, keep growing ranges as new parts of the image are loaded
+  colorRangesTouched = new Map()
```
thewtex, line 49: Perhaps name `colorRangesAdjustWithROI` or similar. For very large datasets, the color range of the largest scale is often much smaller than the true range in the data. This can lead to unexpected / undesirable behavior. In these cases, the colorRange of the full-size, full-resolution dataset can be stored as metadata with the image. 

My impression, please clarify in the comments, is that the map is from image intensity component to a boolean.

src/IO/ConglomerateMultiscaleSpatialImage.js
```diff
@@ -21,11 +21,18 @@ export class ConglomerateMultiscaleSpatialImage extends MultiscaleSpatialImage {
     this.images = images
   }
 
-  async buildImage(scale, bounds) {
-    // Run sequentially rather than Promise.all to avoid hang
+  async getImage(scale, worldBounds = []) {
+    this.worldBoundsStashedForOurBuildImage = worldBounds
```
thewtex, line 25: Perhaps `worldBoundsForBuildImage` to be more succinct.

src/Rendering/Images/createImageRenderingActor.js
```diff
@@ -175,7 +219,10 @@ const eventResponses = {
   // updateRenderedImage->applyRenderedImage->updateCroppingParametersFromImage->CROPPING_PLANES_CHANGED
   // because image size may change across scales.
   CROPPING_PLANES_CHANGED_BY_USER: {
-    target: 'imageBoundsDebouncing',
+    target: 'debouncingUpdatingImage',
+  },
+  CAMERA_MODIFIED: {
+    target: 'debouncingUpdatingImage',
```
thewtex, line 225: debouncingUpdatingImage -> debouncedImageUpdate (remove passive voice)

src/Rendering/VTKJS/Images/applyRenderedImage.js
```diff
@@ -244,86 +281,78 @@ function applyRenderedImage(context, { data: { name } }) {
...
+
+        if (hasChanged) {
           context.service.send({
             type: 'IMAGE_COLOR_RANGE_CHANGED',
-            data: { name, component: componentIndex, range },
+            data: {
+              name,
+              component: componentIndex,
+              range: newRange,
+              fromAutomation: true,
```
thewtex, line 350: fromROI ?

thewtex (changes requested):
Very cool!

Some naming suggestions inline.

We have camera modified debouncing, but do we have other criteria to only update the loaded by a "significant" change in the volume defined by the cropping planes?

src/Context/ImageActorContext.js
PaulHax: Changed to `colorRangesAutoAdjust` as range can expand automatically on scale changes.

src/Rendering/Images/createImageRenderingActor.js
PaulHax: Was going with a convention I read that state are "points in time" rather than functions, so have been throwing a bunch of `ing` endings around.

src/Rendering/VTKJS/Images/applyRenderedImage.js
PaulHax: Now `keepAutoAdjusting`.  Yes, this grow bounds if user has not messed with them logic is confusing.  Not sure how to improve...

PaulHax:
> We have camera modified debouncing, but do we have other criteria to only update the loaded by a "significant" change in the volume defined by the cropping planes?

The cropping planes trigger an image update after being debounced, same debouncer as camera modified, 500ms.  Is that what you mean?

src/Rendering/Images/createImageRenderingActor.js
thewtex: Ah, to satisify the "points in time" and active voice desirables, how about `imageUpdateDebounced`?

Is it possible to not have this state at all and get the functionality with a [condition](https://xstate.js.org/docs/guides/guards.html)?

thewtex:
> The cropping planes trigger an image update after being debounced, same debouncer as camera modified, 500ms. Is that what you mean?

I think what we want is to only update the image content if the component chunks or scale would chancge, e.g. ideally this logic:

1. cropping places or camera changes
2. after debounce
3. for the bounds, create a set of the indexes of the chunks that would be loaded and the required scale
4. add a condition that a hash of the set has changed
5. if the condition passes, go to the state for updating image state (still used that 'ing :-P)

src/Rendering/Images/createImageRenderingActor.js
PaulHax: How would the condition work?  Something like
```
CAMERA_MODIFIED: {
  actions: [assignDebounceTime, c => setTimeout(() => c.server.send('UPDATE_RENDERED_IMAGE_DEBOUCE'), 500)]
  }
UPDATE_RENDERED_IMAGE_DEBOUCE: {
    cond: isAfterDebouceTime, 
    target: 'updatingImage',
}
```
Seems like the declarative nature of a state is preferred over messing with the context.

PaulHax:
> I think what we want is to only update the image content if the component chunks or scale would chancge, e.g. ideally this logic:
> 
> 1. cropping places or camera changes
> 2. after debounce
> 3. for the bounds, create a set of the indexes of the chunks that would be loaded and the required scale
> 4. add a condition that a hash of the set has changed
> 5. if the condition passes, go to the state for updating image state (still used that 'ing :-P)

isImageUpdateNeeded condition does something similar.  Hashes on image IJK indexes, hashing on chunks would have been better.
https://github.com/Kitware/itk-vtk-viewer/blob/a5d40a08c95231c34943fa7d8d51cb8a979dfb79/src/Rendering/VTKJS/Images/imagesRenderingMachineOptions.js#L100-L104

https://github.com/Kitware/itk-vtk-viewer/blob/a5d40a08c95231c34943fa7d8d51cb8a979dfb79/src/Rendering/Images/createImageRenderingActor.js#L190

src/Rendering/Images/createImageRenderingActor.js
thewtex: Yes, that may be roughly be how to implement it, or maybe a boolean instead of a time in the context?

We do want a declarative state, and  we also want it to be valid in its serialized form. If we stop the machine, serialize it, restart it, will debouncing related entries be valid?

thewtex (approved):

PR #565, opened 2022-09-13
fix(setimage): prevent image name from being set to undefined

bnmajor:
If `setImage` is called with no previous data set the image name is `undefined` which creates errors further down the pipeline. Always use a string as the image name.

### DIFF
--- a/src/createViewer.js
+++ b/src/createViewer.js
@@ -702,9 +702,7 @@ const createViewer = async (
   }
 
   publicAPI.setImage = async (image, name) => {
-    if (typeof name === 'undefined' && context.images.selectedName) {
-      name = context.images.selectedName
-    }
+    name = name ?? context.images?.selectedName ?? 'Image'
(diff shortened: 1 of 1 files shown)


src/createViewer.js
```diff
@@ -702,8 +702,8 @@ const createViewer = async (
   }
 
   publicAPI.setImage = async (image, name) => {
-    if (typeof name === 'undefined' && context.images.selectedName) {
-      name = context.images.selectedName
+    if (typeof name === 'undefined') {
+      name = context.images?.selectedName || 'Image'
```
PaulHax, line 706: Cool!  What do you think of the null coalesceing thing?
```
const foundName =  name ?? context.images?.selectedName ?? 'Image'
...
multiscaleImage.name = foundName
```

bnmajor: I love a nice one-liner, thanks for the suggestion!

thewtex (approved):
Woah~~!!?? :-)

PR #568, opened 2022-09-16
fix(ImageRenderingActor): load all image parts in view

PaulHax:
Before if camera moved during loading, machine would not trigger another update, and portions of image in view may not get loaded.

Now, if camera moves during update, toggle loopback child state. When update done and in loopback state, update again.  Guard in update service will stop useless update.


src/Rendering/Images/createImageRenderingActor.js
```diff
@@ -213,18 +213,25 @@ const eventResponses = {
...
-    target: 'debouncedImageUpdate',
+  RENDERED_BOUNDS_CHANGED: {
+    target: 'updatingImage',
   },
 }
 
+const CHANGE_BOUNDS_EVENTS = [
+  'CROPPING_PLANES_CHANGED_BY_USER',
+  'CAMERA_MODIFIED',
+]
```
thewtex, line 224: :+1:

thewtex (approved):

PR #570, opened 2022-09-16
If image loaded after labelImage, reuse actorContext

PaulHax:
Fixes case when setImage and setLabelImage happen after createViewer
```
const viewer = await itkVtkViewer.createViewer(container)
viewer.setImage(image)
viewer.setLabelImage(labelImage)
```


thewtex (approved):
Thanks for adding a test :+1:

PR #573, opened 2022-09-21
Laggy Slice Slider Fix and Out of Memory Catching

PaulHax:
After adding this line
https://github.com/Kitware/itk-vtk-viewer/blob/67915784ef42d03e21c5c8e4b69f980352e7990f/src/Rendering/VTKJS/Images/applyRenderedImage.js#L134
Slider steps were updated every scale to the image spacing.  Synchronous window.render calls became bottlenecks.  This PR changes to asynchronous render calls.

Also catches more out of memory errors when loading images.


thewtex (approved):
Thanks @PaulHax !

PR #577, opened 2022-09-28
Throttle render

### LINKED ISSUE #576: Optimize render calls
Get rid of RENDER_LATER and debounce in response to RENDER when not animating:

https://github.com/Kitware/itk-vtk-viewer/blob/ab848acc7f9127e4623cd7ce395314d160e2fb28/src/Rendering/VTKJS/render.js

PaulHax:
Removes RENDER_LATER.  Adds throttling logic to RENDER:
1. `send("RENER")` from somewhere
2. RenderingMachine gets event and calls render
3. RenderingMachine waits for requestAnimationFrame and one tick via setTimeout to try to wait until after paint
4.  Any `send("RENER")` from anywhere, before above timeout fires, transitions from state waitingForRequest to renderRequested.
5. Throttle setTimeout fires and if in renderRequested, call render

closes #576

### DIFF
--- a/src/Rendering/VTKJS/Images/applyCinematicChanged.js
+++ b/src/Rendering/VTKJS/Images/applyCinematicChanged.js
@@ -48,5 +48,5 @@ export function applyCinematicChanged(context, { actorContext }) {
     }
   })
 
-  context.service.send('RENDER_LATER')
+  context.service.send('RENDER')
 }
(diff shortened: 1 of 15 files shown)


thewtex (commented):
Beautiful, @PaulHax !! This looks like robust logic that will work across renderer implementations.

Can you please review `render()` calls, i.e. `git grep -i 'render()'` and replace with `set('RENDER')` when appropriate?

thewtex:
@bnmajor FYI, a stretch, but this may impact scale / framerate estimation.

bnmajor:
@thewtex Good to know, worth testing out at least. Thanks!

PaulHax:
> Can you please review render() calls, i.e. git grep -i 'render()' and replace with send('RENDER') when appropriate?

Removed a few renderWindow.render() calls.  Still a few in itkVtkViewProxy and the Mobx geometry/points UI code.

thewtex (approved):

Issue #574, opened 2022-09-22
Cannot find IO for nrrd file


idoRosen25:
im trying to display an `.nrrd` file in the viewer from S3.
i have a link for downloading the file.

when i download and copy the file to my project it load without any issues, how ever loading it directly from the s3 link throws an error from the piepline.worker that it cannot find IO for the file.

any idea or suggestions for the issue? (im trying to find a solution that will not require downloading the file to my server every time i want to display it)

PaulHax:
The file extention, `.nrrd`, is used to find the right "IO" flow.  Lets try having your S3 url end in `/filename.nrrd`

idoRosen25:
i am using the url as such but still doesnt load.. i can even see that it found the right filename.. but still cannot recognize the needed IO.. how can i debug or see some errors on that? i cant seem to intercept it..

PaulHax:
Does the URL end in `.nrrd`?  
Yes: https://dandiarchive.s3.amazonaws.com/3d313fc2-0204-496d-bfa1-5c90951ee640.nrrd
No: https://dandiarchive.s3.amazonaws.com/3d313fc2-0204-496d-bfa1-5c90951ee640

For debugging: here is the place passing a URL to 'createViewer' will reach:
https://github.com/Kitware/itk-vtk-viewer/blob/3e546aaab812c18877d43c163587647ad7eb6155/src/IO/toMultiscaleSpatialImage.js#L69
See it grabbing the last part of the URL after the `/` and treating it as the "filename"?

idoRosen25:
my URL is composed as such: `http://<domain>/fileName.nrrd?<authDataForS3>` (since its a protected library).
this url should download the file from S3 (which works and downloads locally on client) but using it directly doesn't find the needed IO.
how can i overcome this issue? (preferably without changing the URL or server logic)

Jmennius:
> Does the URL end in `.nrrd`? Yes: https://dandiarchive.s3.amazonaws.com/3d313fc2-0204-496d-bfa1-5c90951ee640.nrrd No: https://dandiarchive.s3.amazonaws.com/3d313fc2-0204-496d-bfa1-5c90951ee640
> 
> For debugging: here is the place passing a URL to 'createViewer' will reach:
> 
> https://github.com/Kitware/itk-vtk-viewer/blob/3e546aaab812c18877d43c163587647ad7eb6155/src/IO/toMultiscaleSpatialImage.js#L69
> 
[...]

PR #578, opened 2022-09-29
Handle URLs with options and refactor fetchBinaryContent API

Jmennius:
In case URL contains options (common for S3 presigned URLs for example),
the code would not be able to extract filename correctly
and 'IO' flow selection would not work.
Fix this by using URL API to robustly parse filename.

Also, change `fetchBinaryContent()` to work with URL objects instead of strings.
URL objects are usually used in the calling code anyway,
so it makes sense to just pass them.

### DIFF
--- a/src/IO/fetchBinaryContent.js
+++ b/src/IO/fetchBinaryContent.js
@@ -1,8 +1,7 @@
 import axios from 'axios'
 import any from 'promise.any'
 
-async function fetchBinaryContent(url, progressCallback) {
-  const urlObj = new URL(url, document.location)
+async function fetchBinaryContent(urlObj, progressCallback) {
   if (urlObj.protocol === 'ipfs:') {
     const splitPathname = urlObj.href.split('/')
     const cid = splitPathname[2]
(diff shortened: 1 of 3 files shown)


thewtex:
@Jmennius thanks for the contribution :pray: 

Do the tests pass locally for you?

Jmennius:
> @Jmennius thanks for the contribution pray
> 
> Do the tests pass locally for you?

Not really :(
I was struggling to run before submitting (that w3 download..), but i now see how CI gets the data.

Jmennius:
So it's probably because in CI we have Chrome 80.
`Array`'s `at()` is only supported since Chrome 92 [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#browser_compatibility).
I've tried replacing `.at(-1)` with `pop()` and tests pass that way.
I guess we can upgrade Chrome in CI to something recent or use `pop()` or something else.

PaulHax (commented):
`pop()` sounds simple.  Also have another place we need to pass `URL` to `fetchBinaryContent`.  See https://github.com/Jmennius/itk-vtk-viewer/pull/1

PaulHax:
@idoRosen25 Does this fix your #574 issue?

idoRosen25:
> @idoRosen25 Does this fix your #574 issue?

no. the issue was due to the fact that the end of the URL was `fileName.nrrd?queryParams`. adding `.split("?")[0]` did fix the display issue but for some reason made some functions in the event handlers go missing (error of `.someFunc is not a function`). i will test and let you know what i find next week.

PaulHax (approved):
LGTM, thank you for your expertise.

PR #579, opened 2022-09-29
fix(zarrmultiscalespatialimage): use a promise queue to manage resolving chunk promises

bnmajor:
Prevents the network from getting overwhelmed trying to handle all of the promises at once for higher resolution levels. For larger data there may be quite a few chunks and if overwhelmed the data may not load the full resolution.

### DIFF
--- a/src/IO/ZarrMultiscaleSpatialImage.js
+++ b/src/IO/ZarrMultiscaleSpatialImage.js
@@ -7,6 +7,8 @@ import HttpStore from './HttpStore'
 import { CXYZT, toDimensionMap } from './dimensionUtils'
 import { getComponentType } from './dtypeUtils'
 
+import PQueue from 'p-queue'
+
 // ends with zarr and optional nested image name like foo.zarr/image1
 export const isZarr = url => /zarr((\/)[\w-]+\/?)?$/.test(url)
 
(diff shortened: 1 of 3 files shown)


bnmajor:
@PaulHax FYI the promise queue ended up being the solution, thanks for your help!

thewtex (commented):
wOOt!

src/IO/ZarrMultiscaleSpatialImage.js
```diff
@@ -204,6 +206,7 @@ class ZarrMultiscaleSpatialImage extends MultiscaleSpatialImage {
     const chunkPathBase = info.pixelArrayPath
     const chunkPaths = []
     const chunkPromises = []
+    const chunksQueue = new PQueue({ concurrency: 5 })
```
thewtex, line 209: Can we bump this up to something higher, e.g. `10`, `16`, `32`?

bnmajor: Oof, yes, that was a mistake. I actually had it at 10 but some more testing looks like we can set it to 16 with consistent results. Values above that are inconsistent. Fixed to be 16 now.

thewtex (approved):

PR #594, opened 2022-10-06
fix(sidebar): scroll bar shows with small height

PaulHax:
For Reference-UI: moves ColorMapIconSelector popout box left and down to avoid getting clipped when it extends over the right boundary of the sidebar. Also, have to have right side tooltips pop left.  Not ideal.


thewtex (approved):
To the ideal we go :new_moon_with_face:

Issue #590, opened 2022-10-05
RangeError in tests with VTK.js version 25.7.0


PaulHax:
Error in tests:
```
Chrome 80.0.3987.0 (Linux x86_64) LOG: 'ok 77 setImageScale did not error'
Chrome 80.0.3987.0 (Linux x86_64) WARN: 'Resetting view-up since view plane normal is parallel'
Chrome 80.0.3987.0 (Linux x86_64) WARN: 'Resetting view-up since view plane normal is parallel'
Chrome 80.0.3987.0 (Linux x86_64) WARN: 'Resetting view-up since view plane normal is parallel'
Chrome 80.0.3987.0 (Linux x86_64) ERROR: '[mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[Reaction@42]'', RangeError: Invalid typed array length: 16
RangeError: Invalid typed array length: 16
    at new Float32Array (<anonymous>)
    at Object.publicAPI.buildArrays (webpack-internal:///./node_modules/vtk.js/Sources/Rendering/Core/Glyph3DMapper/index.js:179:17)
    at Object.publicAPI.getBounds (webpack-internal:///./node_modules/vtk.js/Sources/Rendering/Core/Glyph3DMapper/index.js:129:15)
    at Object.publicAPI.getBounds (webpack-internal:///./node_modules/vtk.js/Sources/Rendering/Core/Actor/index.js:92:28)
    at handleProp (webpack-internal:///./node_modules/vtk.js/Sources/Proxy/Core/AbstractRepresentationProxy/index.js:311:112)
    at Array.forEach (<anonymous>)
    at Object.publicAPI.getBounds (webpack-internal:///./node_modules/vtk.js/Sources/Proxy/Core/AbstractRepresentationProxy/index.js:325:20)
    at eval (webpack-internal:///./src/Rendering/VTKJS/vtk/ItkVtkViewProxy.js:296:323)
    at Array.forEach (<anonymous>)
    at updateAxes (webpack-internal:///./src/Rendering/VTKJS/vtk/ItkVtkViewProxy.js:295:27)
```
But tests continue without failing anyway

Error starts when using VTK.js version 25.7.0

PR #595, opened 2022-10-07
fix(sidebar): allow clicking under sidebar to 3D canvas

PaulHax:
Also:
* fix clicking logo for credits and FPS
* fix error setting up pointsets in tests.  Gets one of the cases of #590


thewtex (approved):
Thanks, @PaulHax !

PR #599, opened 2022-10-08
fix(createInterface): put color map selector above sliders

### LINKED ISSUE #598: Label image z-index issues
The label image UI is on top of the color map icon selector:

image

@PaulHax any ideas?

PaulHax:
Add 100% height wrapper for uiContainer with pointer-events: none; Set z-index on uiContainer, rather than all children.

Prettify formatting for css.

fixes #598


thewtex (approved):
@PaulHax thanks!

PR #603, opened 2022-10-10
Default colormap

thewtex:


thewtex:
@aylward thanks for the suggestion!

It turns out that CT-Chest-Vessels is similar to CT-AAA, but more linear and with better contrast, so going with for better universality.

bnmajor:
Looks really nice! I am seeing that that colormap selector does not seem to accurately reflect the new default though
cmap_dropdown_bug

thewtex:
@bnmajor thanks for testing! Did you `npm ci` locally before building? There are new packages that may be missing for the UI colormaps.

thewtex:
@bnmajor also `npm run build` or `npm run build:ui` to build the UI package :-)

bnmajor (approved):
@thewtex my mistake - I was using the boostrap-ui without thinking! The reference-ui looks great!

PaulHax (approved):
LGTM!

PR #604, opened 2022-10-10
feat(scaleSelector): spinner shows when loading image

### LINKED ISSUE #596: Indicate when the viewer is changing scales
It would be helpful to have feedback when the viewer is updating its resolution scales. Perhaps pulse the opacity of the scales icon when updating?

PaulHax:
closes #596


thewtex:
@PaulHax getting there!

Naming nitpick: can the events be `IMAGE_UPDATE_STARTED` and `IMAGE_UPDATE_FINISHED` to be more explicit and avoid passive voice?

Testing locally with [this file](https://data.kitware.com/#item/63476e0911dab81428208e27) I only see a flash of the progress icon during updates. Perhaps the events need to be scoped even further out?

thewtex:
Re the IDR use case discussed: https://github.com/InsightSoftwareConsortium/itkwidgets/issues/548 We could also turn the picture icon next to the Image selector into a spinner as the image is loading.

PaulHax:
Tink we got it by waiting to turn off spinner after first render.  `renderWindow.render()` is ignored by vtk.js if interactor is animating so the slow load texture on GPU part was happening later when rotate=true.

Do we want the spinner next to the picture icon?

thewtex:
@PaulHax very nice!!

> Do we want the spinner next to the picture icon?

That would be excellent

PaulHax:
Spinner moved.

src/UI/reference-ui/src/Layers/layersUIMachineOptions.js
```diff
@@ -2,15 +2,16 @@ import createLayersInterface from './createLayersInterface'
...
 const layersUIMachineOptions = {
   layerUIActor: {
     actions: {
       createLayerInterface,
-
       selectLayer,
-
       toggleLayerVisibility,
+      imageUpdating,
+      idle,
```
thewtex, line 14: To enable the same behavior for point sets, geometry, can the `imageUpdating` and `idle` be updated to `startDataUpdate`, `finishDataUpdate`? Internally switch on the actor `type` like `toggleLayerVisibility`.

thewtex (changes requested):
Working wonderfully locally.

A few more tweaks noted inline to pave the way for other datasets.

thewtex (approved):
:100:

PR #605, opened 2022-10-13
feat(createLayersUIMachine): for 4 or more components colormap RGB+Gray

PaulHax:
Set first 4 component colormaps to red, green, blue, gray instead of CT-AAA when more than 4 components

http://localhost:8082/?image=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr

### DIFF
--- a/src/UI/Layers/createLayersUIMachine.js
+++ b/src/UI/Layers/createLayersUIMachine.js
@@ -163,7 +163,7 @@ const assignImageContext = assign({
               colorMap = 'BkBu'
               break
           }
-        } else if (components === 4) {
+        } else if (components >= 4) {
           switch (component) {
             case 0:
               colorMap = 'BkRd'


thewtex:
> something like

This logic could go into a re-usable function(s) in `itk-viewer-color-maps`.  A function that returns a color map for a color map name (internally it checks if the name is a categorical color and generates the cmap for the component), otherwise does the continuous color map Map lookup. And another function that generates the black -> indexedColor continuous colormap from the categorical color name and component.

PaulHax:
Makes sense!  Wonder if folks will make the leap that clicking "glasby/etc" will pick a color from it by component index.  Perhaps another PR for this?  

For quick fix for 5+ component case, loop over RGB+Gray for all higher components' default?

thewtex:
Yes, we can make these improvements in the next iteration -- track in #607

thewtex (approved):

PR #611, opened 2022-10-24
feat(createViewer): disable 2D interpolation if label image

### LINKED ISSUE #610: If a label_image is loaded, Interpolation should be set to False.
If interpolation is applied to a label_image, then values not in the label_map appear in the visualization - making labels appear to be incorrect or undefined at their edges.

From https://github.com/InsightSoftwareConsortium/itkwidgets/issues/565

PaulHax:
closes #610

### DIFF
--- a/doc/content/api/index.md
+++ b/doc/content/api/index.md
@@ -1,5 +1,4 @@
-title: API
----
+## title: API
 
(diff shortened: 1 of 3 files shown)


thewtex (approved):
Thanks @PaulHax !

PR #612, opened 2022-10-27
Material Web for UI

PaulHax:
- Adds Material-Web and Lit as dependencies.  
- Puts sidebar menu in drawer and collapse button to the right.

More fixes todo before good to go.


src/UI/reference-ui/src/context.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const setContext = (providerElement: Element, context: AppContext) => {
+  providerElement.addEventListener('request', event => {
+    const { detail } = (event as unknown) as GetContextEvent
+    detail.context = context
+    event.stopPropagation()
+  })
+}
+
+export function InjectContext(): (target: any, name: string) => any {
```
thewtex, line 17: wow, cool!

src/UI/reference-ui/src/createInterface.js
```diff
@@ -1,13 +1,22 @@
 import style from './ItkVtkViewer.module.css'
+import '@material/web/navigationdrawer/navigation-drawer.js'
 
-import createCollapseUIButton from './createCollapseUIButton'
+import './collapse-ui'
+import { setContext } from './context'
+
+const makeHtml = htmlString => {
```
thewtex, line 7: Nice!!!

thewtex (approved):
@PaulHax brilliant!

I tested locally, and I noticed the menu icon is to the right. I like it.

itsmebasti:
@PaulHax I stumbled upon this project while searching for [@material/web implementations](https://www.npmjs.com/package/@material/web?activeTab=dependents). Yours is the only official project referencing the material web project, so I would like to know where you started from since it is hard to find any documentation. Were you already confident working with material on other Projects, or do you have any handy documentation where I could start working?
I am able to define and use the components but I would love to see any kind of API reference to be aware of the possible component attributes without diving into every single typescript definition.
btw they just released a pre release v1.0, but without any further advice.

Sorry for running into you through this project, there is just barely any other example out there and I am a big fan of web components and would love to use the md components for future projects.
Thanks in advance, Basti

PaulHax:
Hi @itsmebasti I ended up spelunking through the source code, like you suggest, and looking at the Material 2 Web Components examples.  Was not the greatest experience.  I had classic case of shiny new library sickness. Suggest going with something else (Shoelace?) if you can't wait months.

itsmebasti:
Thank you @PaulHax, and thanks for the fast response! Alright, I'll wait for the official md3 statement and go with something else so far. I will have a look into Shoelace, thank you!

PR #613, opened 2022-11-03
build(dependencies): bump vtk.js 25.8.2->25.12.0

bnmajor:


PaulHax:
I took a look at this too, as there are some fixes that would help the viewer...  But there is a quite noticeable performance drop with the latest due to this: https://github.com/Kitware/vtk-js/issues/2618

So unless there is a show stopper maybe good to not degrade the experience with an upgrade just yet?

bnmajor:
@PaulHax Thanks! Makes sense to me, hate to see a slowdown... @thewtex any objections to holding off on the package bump?

thewtex:
:+1: for holding off, thanks @PaulHax @bnmajor :pray:

PR #632, opened 2023-01-13
Resample Wasm

PaulHax:
Firefox does not like itk-wasm's `loadEmscriptenModuleMainThread` function getting called in web worker.  Could modify itk-wasm runPipeline to accept statically imported module?


src/IO/Resample/Resample.cxx
```diff
@@ -0,0 +1,160 @@
...
+
+    ITK_WASM_PRE_PARSE(pipeline);
+
+    return ResampleLabelImage<ImageType>(pipeline, inputImage);
+  }
+};
+
+int main(int argc, char *argv[])
+{
+  itk::wasm::Pipeline pipeline("Resample", "Resample a label image", argc, argv);
```
thewtex, line 147: To be specific `Resample` -> `ResampleLabelImage`

src/IO/Resample/Resample.cxx
```diff
@@ -0,0 +1,160 @@
...
+
+  using ResampleFilterType = itk::ResampleImageFilter<ImageType, ImageType>;
+  auto resampleFilter = ResampleFilterType::New();
+
+  using InterpolatorType = itk::LabelImageGenericInterpolateImageFunction<ImageType, itk::LinearInterpolateImageFunction>;
+  auto interpolator = InterpolatorType::New();
+  resampleFilter->SetInterpolator(interpolator);
+
+  resampleFilter->SetInput(inImage);
+  resampleFilter->SetOutputParametersFromImage(inImage);
```
thewtex, line 65: We want to specify the resampling grid parameters as arguments, i.e. `origin`, `spacing`, `direction`, `size`, and get these from the fixed image / intensity image, example:

https://examples.itk.org/src/filtering/imagegrid/resampleanimage/documentation?highlight=resampleimagefilter

thewtex (commented):
@PaulHax awesome!!

> Firefox does not like itk-wasm's loadEmscriptenModuleMainThread function getting called in web worker. Could modify itk-wasm runPipeline to accept statically imported module?

Follow-up here: https://github.com/InsightSoftwareConsortium/itk-wasm/issues/714#issuecomment-1384734231

src/IO/Resample/Resample.cxx
PaulHax: Feeding those 4 grid parameter to ResampleImageFilter now.

src/IO/ResampleLabelImage/resampleLabelImage.js
```diff
@@ -0,0 +1,58 @@
...
+    new URL(pipelinesWorkerUrlNoBlob, self.location.origin) // URL is invalid for tests unless base provided to URL
+  )
+
+  const { outputs } = await runPipeline(
+    worker,
+    pipeline,
+    taskArgs,
+    desiredOutputs,
+    inputs,
+    pipelinesUrlNoBlob
```
thewtex, line 37: We could not pass this argument -- since the wasm module will be available in the `itkConfig.pipelinesUrl` in this case. The `pipelinesUrl` argument is available as an alternative override.

src/IO/ResampleLabelImage/ResampleLabelImage.cxx
```diff
@@ -0,0 +1,173 @@
+#include "itkBinShrinkImageFilter.h"
```
thewtex, line 1: Add standard ITK Apache 2.0 license header

src/IO/ResampleLabelImage/ResampleLabelImage.cxx
```diff
@@ -0,0 +1,173 @@
...
+  pipeline.add_option("-z,--size", outSize, "New image size for each direction")->expected(2, 3)->delimiter(',');
+
+  std::vector<double> outSpacing;
+  pipeline.add_option("-p,--spacing", outSpacing, "New image spacing for each direction")->expected(2, 3)->delimiter(',');
+
+  std::vector<double> outOrigin;
+  pipeline.add_option("-o,--origin", outOrigin, "New image origin for each direction")->expected(2, 3)->delimiter(',');
+
+  std::vector<double> outDirection;
+  pipeline.add_option("-d,--direction", outDirection, "New image direction")->expected(4, 9)->delimiter(',');
```
thewtex, line 51: Excellent!

src/IO/ResampleLabelImage/index.mjs
```diff
@@ -0,0 +1,77 @@
...
+    '0',
+    '0',
+    '--size',
+    size.join(','),
+    '--spacing',
+    spacing.join(','),
+    '--origin',
+    origin.join(','),
+    '--direction',
+    direction.join(','),
```
thewtex, line 57: Nice

src/IO/ResampleLabelImage/resampleLabelImage.js
```diff
@@ -0,0 +1,58 @@
...
+  const pipelinesWorkerUrlNoBlob = pipelineWorkerUrl.startsWith('blob:')
+    ? pipelineWorkerUrl.substring(5)
+    : pipelineWorkerUrl
+
+  const worker = new Worker(
+    new URL(pipelinesWorkerUrlNoBlob, self.location.origin) // URL is invalid for tests unless base provided to URL
+  )
+
+  const { outputs } = await runPipeline(
+    worker,
```
thewtex, line 32: Using `null` here will generate a worker using the configuration from `itkConfig`.

thewtex (commented):
Looking awesome.

Simplification of asset fetching suggested inline.

src/IO/ResampleLabelImage/resampleLabelImage.js
PaulHax: For the worker, the `__webpack_public_path__` starts with `blob:` for something like `blob:http://...`.  So now overriding `itkConfig.pipelinesUrl` and `itkConfig.pipelineWorkerUrl` after stripping `blob:` off.  

For `npm run test` to work, now composing the relative `/base/dist/...` in the itkConfig with the `self.location.origin` avoids a XMLRequest error about an incomplete URL.

Now can use `null` in runPipeline and leave `pipelinesUrl` parameter off.  This makes WorkerPool happy too so running the split logic now too.

thewtex (approved):
LGTM!

PR #640, opened 2023-01-19
fix(ZarrMultiscaleSpatialImage): Constrain concurrency for large server

thewtex:
While we want to constrain the max concurrency when there are many more logical processors on the client than on the server, we should also do the same when there are many more logical processors on the server than the client.


PaulHax (commented):
I opened a PR pointing to `thewtex:limit-max-concurrency` which is I think what we are going for.

thewtex:
Yes, thank you so much @PaulHax !! :pray: We would not have gotten very far with a concurrency of 0 :snail:

PR #641, opened 2023-01-22
feat(Checkerboard): add Wasm pipeline to process 2 Images

PaulHax:
Adds Viewer function `setCompareImages`

Also adds URL parameters:
`fixedImage` `compare` `pattern` `swapImageOrder`

URL example:
http://localhost:8082/?rotate=false&image=http://localhost:8082/test-data/HeadMRVolume.nrrd&fixedImage=http://localhost:8082/test-data/HeadMRVolume2Components.nrrd&compare=checkerboard&pattern=[1,3,4]&swapImageOrder=true

Zarr URL test:
http://localhost:8082/?rotate=false&image=http://localhost:8082/test-data/idr/6001240.zarr&fixedImage=http://localhost:8082/test-data/idr/6001256.zarr&compare=checkerboard&pattern=[3,3,4]&swapImageOrder=true

Rebuilds Downsample and ResampleLabelImage Wasm binaries with itk-wasm 1.0.0.-b.84

TODO
- [x] Feed boxes pattern parameter through with undefined default for filter's default choice
- [x] Fix viewer interface, state handling, and pipeline processing for `fixedImage` 
- [x] Add test(s)
- [x] Docs
- [x] Factor out runWasm()
- [x] URL parameters?
- [x] How to deal with multi-component images? Answer: RGBToLuminanceImageFilter or VectorMagnitudeImageFilter
- [x] Resample fixed to moving if needed
- [x] Add test with multichannel input
- [x] Fuse checkerboard output with label map component
- [x] Test with Conglomerate
- [x] Update `visualizedComponents` and UI when we compress down to 1 component.
- [x] test with 2d, conglomerate
- [x] Change checkerboard to lowercase.
- [x] Squash
- [x] ~~Remove `*.wasm`'s.  Kept `.umd.wasm`'s~~ Need `umd.wasm` for local runtime and `.wasm` for tests in CI!?!?
- [x] Ensure CI tests are passing
- [x] Merge!


thewtex:
- ComposeImageFilter -> stack for cyan/magenta
- Apply ResampleImageFilter to sample on the fixed image
- RGBToLuminanceImageFilter or VectorMagnitudeImageFilter
- Add test with multichannel input

PaulHax:
How the checkerboard flow will work?
image

thewtex:
@PaulHax looks excellent!!

A few clarifications:

- "Resample fixed to moving" -> "Resample moving to fixed" (move to other path -- operates on moving as the input). This avoid "holes" in the resampling process.

- *Map Fixed into type space of Moving* -> *Map Moving into type space of Fixed*. We are already resampling/processing the moving, so we can also convert its type in the same pipeline.

- *XXX or RescaleIntensityImageFilter* -> *XXX (IntensityWindowingImageFilter + MinimumMaximumImageCalculator) or RescaleIntensityImageFilter*

- In the last pipeline, this could be A) Checkerboard, but also options for B) Compose, or C) just Resample.

PaulHax:
@thewtex How do I pass 2 images of different types to a pipeline?  I posted on the fourm:
https://discourse.itk.org/t/itk-wasm-pipelinefunctor-with-2-input-image-types/5657

I could also just do the type conversion and Rescaling in Javascrpt.

thewtex:
> How do I pass 2 images of different types to a pipeline?

@PaulHax example here:

https://github.com/InsightSoftwareConsortium/itk-wasm/pull/746

However, we should use `itk-wasm` `castImage` as needed in pre-processing to keep the size of the Wasm binary down so only one set of input image types need to be supported.

> Another question: How to conditionally run VectorMagnitudeImageFilter and wire it into the filter flow. I'm not sure how to line up my types: Somehow extract the Type out of Image<Vector<Type>> and put in

The logic should likely use a runtime check for the number of components in the image versus a compile time check. This can be checked via [GetNumberOfComponentsPerPixel](https://itk.org/Doxygen/html/classitk_1_1Image.html#a18474315e7f0bc39e4142a44a209d514).

thewtex (commented):
👏

PR #659, opened 2023-03-15
feat(compareUI): add inputs for checkerboard pattern and swap image

PaulHax:
image

Still need a way to turn off compare method.


thewtex (approved):
:sparkler:

PR #661, opened 2023-03-19
close all workers when process is done

jonSeetrue:
Fix the memory leak issue.
close all workers when done processing.

### DIFF
--- a/src/Compression/bloscZarrDecompress.js
+++ b/src/Compression/bloscZarrDecompress.js
@@ -4,6 +4,7 @@ import { getSize } from '../IO/dtypeUtils'
 const cores = navigator.hardwareConcurrency ? navigator.hardwareConcurrency : 4
 const numberOfWorkers = cores + Math.floor(Math.sqrt(cores))
 const workerPool = new WorkerPool(numberOfWorkers, runPipeline)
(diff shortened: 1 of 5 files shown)


jonSeetrue:
Hi @banesullivan 
  This pr fixes the memory issues of the open workers, could you review it, please?

thewtex (changes requested):
@jonSeetrue yes, thank you for the contribution!

Can you please call `terminateWorkers()` from a `cleanup()` function on the viewer object? For performance, we want to re-use the workers on repeated calls.

jonSeetrue:
@thewtex  do you mean to collect all the workers into an array on the viewer object and when loading a new image call terminateWorkers()?

thewtex:
@jonSeetrue a `cleanup()` function on the viewer that calls `terminateWorkers()` on the worker pools when called. Loading a new image will not need a call. I am not sure if an array on the viewer object is needed.

jonSeetrue:
@thewtex  the worker-pool is not exported, so should the cleanup() function be part of each worker-pool implementation?

thewtex:
@jonSeetrue yes, that could work :+1:

PR #666, opened 2023-04-04
fix(transferfunctionmanipulators)

bnmajor:
Cap the step value for the window/level interactor to 10. The `vtkMouseRangeManipulator` appears unresponsive when the step value >= 100.


thewtex (approved):
:+1:

PR #667, opened 2023-04-06
feat(applycolorrangebounds): do not clamp the color range

bnmajor:


thewtex (approved):
@bnmajor thanks!!

CC @aylward

PR #669, opened 2023-04-11
fix(createViewer): Handle TAKE_SCREENSHOT event

thewtex:
To address with itkwidgets:

```
caught Error: Unexpected event type: TAKE_SCREENSHOT
    at itkVtkViewer.js:161:457719
    at itkVtkViewer.js:2:1470699
    at Set.forEach (<anonymous>)
    at Object.send (itkVtkViewer.js:2:1470671)
    at e.sendTo (itkVtkViewer.js:2:1457639)
    at e._exec (itkVtkViewer.js:2:1458205)
    at e.exec (itkVtkViewer.js:2:1467591)
    at e.execute (itkVtkViewer.js:2:1459688)
    at e.update (itkVtkViewer.js:2:1460057)
    at itkVtkViewer.js:2:1457049
```

### DIFF
--- a/src/createViewer.js
+++ b/src/createViewer.js
@@ -181,6 +181,8 @@ const createViewer = async (
           case 'SCREENSHOT_TAKEN':
             eventEmitter.emit('screenshotTaken', event.data)
             break
+          case 'TAKE_SCREENSHOT':
+            break
           default:
             throw new Error(`Unexpected event type: ${event.type}`)
         }


bnmajor (approved):
Awesome, thank you!

PaulHax (approved):
I remember seeing this error, but does not seem to happen every time.  Wonder where event comes from?

thewtex:
In itkwidgets, we take a screenshot by default after initial rendering and this is saved in the notebook output cell so when browsing on GitHub, etc. there is a rendering in the notebook.

I am wondering if this errors ends putting the viewer in an invalid state. Subsequent view commands can be unresponsive.

PR #671, opened 2023-04-12
Ingestion fixes for 2D images 

PaulHax:
Image would not show up if `direction` matrix was 2D or if a dimension's spatial coordinate array had only 1 element.

### DIFF
--- a/src/IO/MultiscaleSpatialImage.js
+++ b/src/IO/MultiscaleSpatialImage.js
@@ -222,10 +222,9 @@ class MultiscaleSpatialImage {
     const spacing = new Array(this.spatialDims.length)
     for (let index = 0; index < this.spatialDims.length; index++) {
       const dim = this.spatialDims[index]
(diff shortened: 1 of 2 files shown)


src/IO/MultiscaleSpatialImage.js
```diff
@@ -224,7 +224,8 @@ class MultiscaleSpatialImage {
       const dim = this.spatialDims[index]
       if (info.coords.has(dim)) {
         const coords = await info.coords.get(dim)
-        spacing[index] = coords[1] - coords[0]
+        // if has one entry for dim, just use, else find difference between first 2
+        spacing[index] = coords.length === 1 ? coords[0] : coords[1] - coords[0]
```
thewtex, line 228: `coords[0]` is the origin -- to be consistent with other behavior we should use `1.0` as the default if there is not spacing.

PaulHax: Fixed.

PR #685, opened 2023-05-04
feat(updateRenderedImage): cap memory size of input image

PaulHax:


thewtex (approved):
:medal_sports: :1st_place_medal: :100:

PR #688, opened 2023-05-11
Bump vtkjs

bnmajor:
and set rendering backend to WebGL


thewtex (commented):
LGTM! @bnmajor thank you!

@PaulHax is going to review / merge.

PaulHax:
Good stuff.  

Was getting a `structuredClone` not defined error after `npm run test:debug`.  Upgrading puppateer to `"puppeteer": "^20.2.0",` upgrades the Chrome version used in the test and it passes.  

@bnmajor Can you bump puppeteer and and push?

Looks like there is a bug in VTK.js after changing the volume blend mode. https://github.com/Kitware/vtk-js/issues/2830
I'ed vote to not hold this PR for that fix.

bnmajor:
Thanks @PaulHax! Puppeteer has been bumped

PaulHax:
Gaw.  Another ask.  CI test job is failing at the same test as `npm run debug` with the old puppeteer (just after 41 reset cropping planes.)  Can we avoid increasing the vtk.js version and just specify WebGL and does that fix the issue?

bnmajor:
@PaulHax We can skip the vtk.js bump but it doesn't look like it fixes anything - my local tests are still failing [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/RenderWindow/index.js#L287) with the following:
```
caught (in promise) TypeError: Cannot create proxy with a non-object as target or handler
    at publicAPI.get3DContext (index.js:287:12)
    at publicAPI.initialize (index.js:170:33)
    at publicAPI.buildPass (index.js:161:17)
    at publicAPI.apply (index.js:44:7)
    at publicAPI.traverse (index.js:32:15)
    at publicAPI.traverse (index.js:26:14)
    at publicAPI.traverseAllPasses (index.js:1213:35)
    at forceRender (index.js:452:19)
    at publicAPI.render (index.js:899:7)
    at publicAPI.initialize (index.js:148:15)
```

Looks like `model.defaultToWebgl2` and `webgl2Supported` return `true` but `model.canvas.getContext('webgl2'/'webgl'/'experimental-webgl', options)` are all returning null. FWIW this is the same error I am getting with the version bump.

PaulHax:
OK, thanks for checking.

looks like adding a puppeteer config file actually gets the CI tests running.  But they failing for the usual mysterious reasons. I am looking into it.

PR #691, opened 2023-05-15
Small vtkjs bump to 26.9.15

PaulHax:


PaulHax:
@bnmajor I'm having trouble with tests after updating puppeteer.  

There is a back-ported fix for WebGPU for our current vtk.js major version.  Can you quickly test if this PR works, and we can get this fix in?

bnmajor:
@PaulHax This is working for me, thank you for the fix!!

PR #699, opened 2023-05-29
Bbox

bnmajor:
This PR adds a button that allows the user to toggle on/off a label image bounding box. This will be useful when trying to visualize the actual bounds of a label image when it is not rendered as expected. This currently only adds this feature for the label image. This can be changed if the feature should be available for all images (moving, fixed, label, etc).

label_bbox

Currently a draft - Relies on InsightSoftwareConsortium/itk-viewer-icons#5.

Should help with InsightSoftwareConsortium/itkwidgets#663, InsightSoftwareConsortium/itkwidgets#662.


src/Context/LayersMachineContext.js
```diff
@@ -7,6 +7,9 @@ class LayersMachineContext {
 
   // A { name, data } object, queued for creation of the data's actor
   lastAddedData = null
+
+  // Draw bounding boxes around image and label (if loaded)
+  labelBBoxEnabled = false
```
PaulHax, line 12: Should this state be associated with the actual image, not in this image aggregating parent?

src/Rendering/Images/createImagesRenderingMachine.js
```diff
@@ -148,6 +148,11 @@ function createImagesRenderingMachine(options, context) {
                 to: (c, e) => `imageRenderingActor-${e.data}`,
               }),
             },
+            TOGGLE_LABEL_BBOX: {
+              actions: send((_, e) => e, {
+                to: c => `imageRenderingActor-${c.images.selectedName}`,
```
PaulHax, line 153: Should this event include the name of the image in the payload like the the other events?  So the UI layer pulls `selectedName` and puts it in `data: { name: selectedName }`.  Then can toss the event name on the `makeTransitions` `forwardToNamedActor` block below. Things are inconsistent, but better every day.

src/Rendering/VTKJS/vtk/ItkVtkViewProxy.js
```diff
@@ -702,6 +704,16 @@ function ItkVtkViewProxy(publicAPI, model) {
   model.axesCircleRadius = 4
   model.axesTextOffset = 14
 
+  model.enableLabelBBox = false
```
PaulHax, line 707: Is it possible to avoid duplication of state and pull this boolean from the context?

src/createViewer.js
```diff
@@ -883,10 +883,16 @@ const createViewer = async (
...
     const imageName =
       layerImageName ?? context.images.selectedName ?? multiscaleLabelImage.name
-    service.send({
-      type: 'ADD_LABEL_IMAGE',
-      data: { imageName, labelImage: multiscaleLabelImage },
-    })
+    const actorContext = context.images.actorContext.get(imageName)
+    if (actorContext?.labelImageName === multiscaleLabelImage.name) {
+      actorContext.labelImage = multiscaleLabelImage
+      service.send({ type: 'LABEL_IMAGE_ASSIGNED', data: imageName })
```
PaulHax, line 889: Good fix.

src/Context/LayersMachineContext.js
bnmajor: This is now a map of each layer and its bounding box visibility status. That makes sense though. Maybe each layer's status should live with its `LayerActorContext` instead of at the top level here?

src/Rendering/Images/createImagesRenderingMachine.js
bnmajor: Fixed, thanks :+1:

src/Rendering/VTKJS/vtk/ItkVtkViewProxy.js
bnmajor: I am using the model's enabled flag to [guard against](https://github.com/Kitware/itk-vtk-viewer/pull/699/files#diff-ea69b57bef01cbc2983c1fd92fdc8bce80a470cda143616cb12b46643f6c1a71R953-R963) unnecessary render calls by comparing it to the context state.

src/createViewer.js
bnmajor: Thanks!

src/Context/LayersMachineContext.js
PaulHax: Yeah. Or the `ImageActorContext`, i'm not sure.

src/Rendering/VTKJS/vtk/ItkVtkViewProxy.js
PaulHax: Could pull the visibility boolean from the actor. Or just call `actor.setVisiblity` and if vtk.js is working right, it will do the guard check and won't render if nothing changed.

src/Context/LayersMachineContext.js
bnmajor: I've moved it to the `LayerActorContext` for now as each layer can simply have a boolean `bbox` key associated with it. Moving it to the `ImageActorContext` would be just as simple, we'd just need an `imageBBox`  and `labelBBox` (or something along those lines) associated with each image actor.

src/Rendering/VTKJS/vtk/ItkVtkViewProxy.js
bnmajor: Cleaned up :+1:

src/Context/LayersMachineContext.js
thewtex: Most excellent.

In the future, it will also be useful to show the bounding box for point sets, etc.

thewtex:
Awesome!!

PR #723, opened 2023-11-30
feat(createLayerInterface): add download button

PaulHax:


src/viewerMachineOptions.js
```diff
@@ -8,6 +8,17 @@ const ViewerMachineOptions = {
...
     styleRenderingViewContainers,
+    downloadImage: (context, event) => {
+      const { croppingPlanes } = context.main
+      context.service.send({
+        type: 'SAVE_ROI',
+        data: {
+          name: event.data.name,
+          layerName: event.data.layerName,
+          croppingPlanes,
+        },
```
PaulHax, line 19: Thee emited event payload.

src/createViewer.js
```diff
@@ -183,6 +183,9 @@ const createViewer = async (
             break
           case 'TAKE_SCREENSHOT':
             break
+          case 'SAVE_ROI':
+            eventEmitter.emit('saveRoi', event.data)
```
PaulHax, line 187: The event.

thewtex (changes requested):
The workflow for itkwidgets / python side is:

1. create a viewer
2. select an roi
3. there are getters to get the current roi on the viewer
4. the client code calls the getter and their logic to save the image to file

The download button should bring up a modal to select format and file name parameters and a download button that saves the file locally.

We should not be emitting an event / payload for the save ROI.

The download button is not used for this because of a) how event processing occurs in jupyter and b) how the roi is obtained and save logic is implemented in Python.

-> please add the download modal and remove the event

PaulHax:
Closing in favor of #727

PR #728, opened 2023-12-18
feat: add independent GUI change of color range

PaulHax:


thewtex (approved):
Looks great!

thewtex:
:heart_eyes: 

Beautiful!!

Defaults are on point.

image

PR #729, opened 2023-12-18
fix: default label image weight for background label is 0.0

thewtex:
Instead of 0.1.

To address: https://github.com/InsightSoftwareConsortium/itkwidgets/issues/683


PaulHax (approved):
Bada-bing

PR #751, opened 2024-06-21
feat(ImJoyPluginAPI): add setImagePiecewiseFunctionPoints and getter

PaulHax:


src/ImJoyPluginAPI.js
```diff
@@ -253,6 +253,14 @@ class ImJoyPluginAPI {
     return this.viewer.getImagePiecewiseFunctionGaussians(component, name)
```
thewtex, line 253: @PaulHax These are no longer valid? They should be removed?

thewtex (commented):
:

src/ImJoyPluginAPI.js
PaulHax: Yes.  I just removed them.

thewtex (approved):
:rocket:

